PR #28, opened 2016-10-28
Update packages.json with react-addons-css-transition-group.

jasonzoladz:


jasonzoladz:
Cached `-prelude` fouled this up.  Fixing now.

PR #42, opened 2017-01-05
Add eulalie, signal, smolder and vdom.

bodil:
Also, bump test-unit.


paf31:
Thanks!

paf31:
Oops, I just realized the latest CI job is not pulling the package set from the PR branch. I created #43 to track that.

PR #46, opened 2017-01-25
Add fork.

seagreen:


seagreen:
This add v1.0.0 of `fork` ([from Nov. 19](https://github.com/slamdata/purescript-fork/releases)) instead of v1.1.0 (from Nov 22), because it's in preparation for adding halogen to the package set once the new halogen's released, and v.1.0.0 is what [halogen currently requires](https://github.com/slamdata/purescript-halogen/blob/master/bower.json#L33).

paf31:
Thanks!

PR #60, opened 2017-03-29
Some contrib packages

paf31:


paf31:
Obviously these won't be usable until I make a tag for 0.11.2.

PR #73, opened 2017-05-25
Add mmorph, spec, pipes

mostalive:
(Made a mistake - thought I had merged back from purescript/master, but had my origin set wrongly, so remotedata is in here again.)


paf31:
Thanks!

PR #77, opened 2017-06-05
Updates for 0.11.5

paf31:


paf31:
It seems there is some mismatch between `smolder` and `pux`:

```
Error found:
in module Pux.DOM.HTML
at .psc-package/testing/pux/v9.1.0/src/Pux/DOM/HTML.purs line 34, column 31 - line 34, column 32
  Could not match type
        
    Unit
        
  with type
                              
    MarkupM (Event -> a1) Unit
                              
while trying to match type MarkupM t2 Unit
  with type t0 (MarkupM (Event -> a1) Unit)
while checking that expression c
  has type t0 (MarkupM (Event -> a1) Unit)
in binding group mapEvent
where a1 is a rigid type variable
        bound at line 33, column 1 - line 34, column 78
      t2 is an unknown type
      t0 is an unknown type
```

@alexmingoia Should I just downgrade `smolder` for now, or do you have plans to upgrade to the latest version? Either way is fine for now. Thanks!

paf31:
Hmm, `smolder-8.0.0` doesn't seem to work either.

PR #80, opened 2017-06-12
Add dom classy

MichaelXavier:


paf31:
Thanks! And thanks for breaking these up into separate PRs.

PR #81, opened 2017-06-12
Add errors

MichaelXavier:


MichaelXavier:
@paf31 ok, this one's done. After this I'll do #84 and then I should be done for a while.

paf31:
👍

PR #86, opened 2017-06-14
Formatters

MichaelXavier:
Looks like somehow duplicates were in the master so there's a separate commit removing them. I also could not really split this into separate PRs as formatters required fixed-points.


paf31:
Thanks!

PR #90, opened 2017-07-06
More halogen.

krisajenkins:
This PR adds a few packages from pursuit, notably some of the halogen support libraries. Details in each commit.


paf31:
Great, thanks!

PR #99, opened 2017-09-06
add email-validate

tippenein:
https://github.com/cdepillabout/purescript-email-validate/blob/master/package.json


paf31:
Thanks!

PR #101, opened 2017-09-09
Add flare

joneshf:
Part of https://github.com/purescript/trypurescript/issues/103

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -339,6 +339,20 @@
     "repo": "https://github.com/slamdata/purescript-fixed-points.git",
     "version": "v4.0.0"
   },
+  "flare": {
+    "dependencies": [
+      "canvas",
+      "datetime",
+      "dom",
+      "drawing",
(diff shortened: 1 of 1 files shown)


sharkdp:
@joneshf Thank you very much for working on this!


Currently, try-flare also supports purescript-flarecheck (now called [puresript-sparkle](https://github.com/sharkdp/purescript-sparkle)). It would be great if we could still support this.

If I understand this correctly, we would need to add something like this:
``` json
  "sparkle": {
    "dependencies": [
      "flare",
      "proxy",
      "globals",
      "generics",
      "record",
      "typelevel-prelude",
      "numbers"
    ],
    "repo": "https://github.com/sharkdp/purescript-sparkle.git",
    "version": "v4.1.1"
  },
```

paf31:
Looks good, but yes, let's add `sparkle` too (separately).

paf31:
Thanks!

PR #106, opened 2017-09-24
Update some deps and add simple-json

justinwoo:
Please let me know if there's a reason why we shouldn't have some of these though.


paf31:
👍 LGTM, thanks!

PR #107, opened 2017-09-24
add even more packages

justinwoo:
...as used in https://github.com/justinwoo/vidtracker/commit/e9ab99f1193cdcd92049e8855ef948ecc6ab9063


paf31:
Great, thanks!

PR #111, opened 2017-09-29
Put together an Aff 4.0 package set

justinwoo:
Putting this together to test it out, hopefully I can use it as a long-living todolist to try to get the removed packages fixed.

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -1367,6 +1367,24 @@
     "repo": "https://github.com/purescript/purescript-psci-support.git",
     "version": "v3.0.0"
   },
+  "pux": {
+    "dependencies": [
+      "aff",
+      "eff",
+      "lists",
+      "signal",
+      "maps",
+      "functions",
+      "profunctor",
+      "react",
+      "globals",
+      "dom",
+      "smolder",
+      "css"
+    ],
+    "repo": "git://github.com/alexmingoia/purescript-pux.git",
+    "version": "v12.0.0"
+  },
   "quickcheck": {
     "dependencies": [
       "arrays",
--- a/packages.json
+++ b/packages.json
@@ -1080,6 +1080,11 @@
     "repo": "https://github.com/purescript-node/purescript-node-fs-aff.git",
     "version": "v5.0.0"
   },
+  "node-he": {
+    "dependencies": [],
+    "repo": "https://github.com/justinwoo/purescript-node-he.git",
+    "version": "v0.1.0"
+  },
   "node-http": {
     "dependencies": [
       "maps",
--- a/packages.json
+++ b/packages.json
@@ -342,7 +342,7 @@
       "unsafe-coerce"
     ],
     "repo": "https://github.com/purescript-contrib/purescript-dom.git",
-    "version": "v4.9.0"
+    "version": "v4.10.0"
   },
   "dom-classy": {
     "dependencies": [
@@ -358,7 +358,7 @@
       "dom"
     ],
     "repo": "https://github.com/slamdata/purescript-dom-indexed.git",
-    "version": "v3.0.0"
+    "version": "v5.0.0"
   },
   "drawing": {
     "dependencies": [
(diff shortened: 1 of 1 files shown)


justinwoo:
Verified the set locally, let's see what Travis has to say about it...

justinwoo:
Reminder to self that io and pipes are updated, so I need to look into those eventually

paf31:
Thermite is updated too.

packages.json
```diff
@@ -321,8 +345,8 @@
        "aff",
        "transformers"
     ],
-    "repo": "https://github.com/cdepillabout/purescript-email-validate.git",
-    "version": "v1.0.7"
+    "repo": "https://github.com/justinwoo/purescript-email-validate.git",
+    "version": "aff-4.0"
```
justinwoo: NOTE TO SELF THAT THIS NEEDS TO BE UPDATED

justinwoo: https://github.com/cdepillabout/purescript-email-validate/pull/5

packages.json
```diff
@@ -795,16 +788,6 @@
     "repo": "https://github.com/paf31/purescript-memoize.git",
     "version": "v4.0.1"
   },
-  "milkis": {
```
justinwoo: depends on aff promise

packages.json
```diff
@@ -1,40 +1,53 @@
...
   },
   "aff-coroutines": {
     "dependencies": [
       "aff",
       "coroutines"
     ],
     "repo": "https://github.com/purescript-contrib/purescript-aff-coroutines.git",
-    "version": "v5.0.0"
-  },
-  "aff-promise": {
```
justinwoo: depends on PR stuff

packages.json
```diff
@@ -570,52 +594,6 @@
     "repo": "https://github.com/purescript/purescript-graphs.git",
     "version": "v3.0.0"
   },
-  "halogen": {
-    "dependencies": [
```
justinwoo: halogen family wip

packages.json
```diff
@@ -878,24 +861,6 @@
     "repo": "https://github.com/purescript-node/purescript-node-path.git",
     "version": "v2.0.0"
   },
-  "node-postgres": {
```
justinwoo: wip

packages.json
```diff
@@ -1106,17 +1077,6 @@
     "repo": "https://github.com/purescript/purescript-psci-support.git",
     "version": "v3.0.0"
   },
-  "pux": {
```
justinwoo: wip

packages.json
```diff
@@ -1244,14 +1194,6 @@
     "repo": "https://github.com/purescript/purescript-semirings.git",
     "version": "v4.0.0"
   },
-  "servant-support": {
```
justinwoo: wip

packages.json
```diff
@@ -1301,23 +1243,19 @@
...
+      "transformers",
+      "monoid",
+      "foldable-traversable",
+      "pipes",
+      "ansi"
     ],
     "repo": "https://github.com/owickstrom/purescript-spec.git",
-    "version": "v1.0.0"
-  },
-  "spec-quickcheck": {
```
justinwoo: wip

packages.json
```diff
@@ -1082,380 +992,362 @@
...
-      "maps"
+      "tuples",
+      "free",
+      "transformers",
+      "bifunctors"
     ],
     "repo": "https://github.com/bodil/purescript-smolder.git",
-    "version": "v9.0.0"
-  },
-  "sparkle": {
```
justinwoo: I don't know how to upgrade this package from Monoid MarkupM to non-monoid so I'm just going to drop this

packages.json
justinwoo: in the end, i couldn't figure out how to make pux actually build correctly in CI. need to experiment further later.

justinwoo:
I think the packages dropped from the set will take a while before they can be updated for use, but the other ones all work fine. Could we merge and tag this, if there aren't any problems?

paf31:
I'm not sure if we want to drop packages between major releases like this, especially something like Halogen.

Could you please make a list of the dropped packages?

packages.json
```diff
@@ -1109,81 +1279,111 @@
...
+      "profunctor",
+      "react",
+      "globals",
+      "dom",
+      "smolder",
+      "css"
     ],
     "repo": "https://github.com/alexmingoia/purescript-pux.git",
-    "version": "v10.0.0"
+    "version": "v12.0.0"
```
justinwoo: This had better work, it worked here: https://travis-ci.org/justinwoo/package-sets/builds/286428685

justinwoo:
Here are the packages that are missing:

* halogen
  WIP PR
* halogen-bootstrap
  depends on Halogen
* halogen-css
  depends on Halogen
* halogen-echarts
  depends on Halogen
* halogen-vdom
  depends on Halogen
* node-postgres
  WIP PR
* sparkle
  No PR
* spec-quickcheck
  WIP PR

There's also something seriously wrong with the travis setup here or something, as I can't get Pux to verify even though it works already for me: https://travis-ci.org/justinwoo/package-sets/builds/286428685, https://travis-ci.org/justinwoo/package-sets/builds/286432483

justinwoo:
I guess at minimum we should wait until Halogen is in, and then the others seem a bit optional whether or not people have actually use them, but maybe Node-Postgres is important to have too.

packages.json
anttih: https://github.com/epost/purescript-node-postgres/releases/tag/v4.0.0

paf31:
I've reset the cache and restarted the build.

paf31:
I'm thinking it would be good to merge this into the 0.11.7 package set, since I'm thinking of making a 0.11.7 release this week to put out some bug fixes and IDE improvements.

paf31:
Thanks very much!

PR #116, opened 2017-10-30
Adds purescript-int-53 and purescript-jack

damncabbage:
Adds [purescript-int-53](https://github.com/rgrempel/purescript-int-53) and [purescript-jack](https://github.com/jystic/purescript-jack).


paf31:
Great, thanks!

PR #119, opened 2017-11-09
Update README.md

### LINKED ISSUE #117: Needs some docs on how to add a package to the set
With the new `verify` command, the workflow for updating and validating a package set is pretty easy, but this needs to actually be written down somewhere how it works. Probably information should live either in this repo or in the section in the README of psc-package.

Anyone with opinions about where it *should* go? I feel like it's easier to read when it's here already while you have this repo checked out to prepare a package set.

justinwoo:
fixes #117


paf31:
Good idea, thanks!

PR #124, opened 2017-11-20
Changing some git-protocol urls to https ones.

krisajenkins:
Sadly, the git-protocol URLs can cause problems for those of us behind nasty corporate
firewalls.


paf31:
Makes sense, thanks! We should probably document this preference somewhere actually.

PR #125, opened 2017-11-20
Restoring Halogen-Echarts.

krisajenkins:
Not sure why, but it got removed in the recent 0.11.7 release. Assuming that was an oversight, here's the fix.


paf31:
Thanks!

PR #126, opened 2017-11-23
update behaviors

justinwoo:
need some updates from this

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -162,7 +162,7 @@
       "nullable"
     ],
     "repo": "https://github.com/paf31/purescript-behaviors.git",
-    "version": "v6.0.1"
+    "version": "v6.2.0"
   },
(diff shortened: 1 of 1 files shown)


packages.json
```diff
@@ -161,8 +161,8 @@
       "filterable",
       "nullable"
     ],
-    "repo": "https://github.com/paf31/purescript-behaviors.git",
-    "version": "v6.0.1"
+    "repo": "git://github.com/paf31/purescript-behaviors.git",
```
kRITZCREEK: Let's prefer https: urls here. Having to ssh can be a problem with Proxys

justinwoo: ah, sorry, that's a bad rebase from me

paf31:
Thanks!

PR #131, opened 2017-12-06
More on aff 4

justinwoo:
just PRing my hodge podge of packages. Could remove some that are more specific to my own needs.

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -214,6 +214,16 @@
     "repo": "https://github.com/purescript/purescript-catenable-lists.git",
     "version": "v4.0.0"
   },
+  "choco-pie": {
+    "dependencies": [
+      "prelude",
+      "behaviors",
+      "typelevel-prelude",
+      "record"
+    ],
+    "repo": "git://github.com/justinwoo/purescript-chocopie.git",
+    "version": "v1.0.0"
+  },
   "codec": {
     "dependencies": [
       "transformers",
--- a/packages.json
+++ b/packages.json
@@ -888,6 +888,21 @@
     "repo": "https://github.com/jystic/purescript-jack.git",
     "version": "v2.0.0"
   },
+  "io": {
+    "dependencies": [
+      "aff",
+      "control",
+      "eff",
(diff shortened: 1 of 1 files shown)


packages.json
```diff
@@ -1926,6 +2003,14 @@
     "repo": "https://github.com/natefaubion/purescript-variant.git",
     "version": "v4.0.0"
   },
+  "xstream": {
+    "dependencies": [
+      "aff-promise",
+      "js-timers"
+    ],
+    "repo": "git://github.com/justinwoo/purescript-xstream.git",
```
paf31: `git` -> `https` here pls :)

packages.json
```diff
@@ -1188,6 +1237,17 @@
...
   },
+  "node-telegram-bot-api": {
+    "dependencies": [
+      "aff",
+      "strings",
+      "foreign",
+      "aff-promise",
+      "simple-json"
+    ],
+    "repo": "git://github.com/justinwoo/purescript-node-telegram-bot-api.git",
```
paf31: Another `git`

oliver3:
Hi Justin,

Very possible that I just don't understand how things work, but to me it seems that `node-telegram-bot-api` v1.0.0 is still depending on purescript-aff ^3.0.0, so I am not sure how this package-set could work... Only way I see is that the 1.0.0 is actually not the same as the "bower version" 1.0.0 ?

justinwoo:
Sorry, commented in the wrong spot at first. Yeah, seems strange to me that I was able to use the existing 1.0.0 in https://github.com/justinwoo/simple-rpc-telegram-bot/commit/c7199ca8d72344d6c5da8a30daa15dd03622c2e0  where the package set I used there very clearly points at 1.0.0, but it still builds fine. But I guess there's no problems since I use a very few of what is there?

oliver3:
Ah thanks @justinwoo, also for your reply on the psc-package issue, and for releasing the telegram API

LiamGoodacre:
Thanks!

PR #133, opened 2017-12-17
Add sparkle back

joncfoo:
It seems this package was accidentally removed as part of #111 (9a5778b2f2).

Found while trying to build all parts of trypurescript.


joncfoo:
Ah so CI says Sparkle is broken and it was probably intentionally removed!

There seems to be a relevant issue filed at https://github.com/sharkdp/purescript-sparkle/issues/12

I'll see if I can tackle the root cause and then circle back to this.

Closing for now.

PR #148, opened 2018-02-15
Add `prettier-printer` v1.2.0

joneshf:
Re: https://github.com/paulyoung/purescript-prettier-printer/issues/9

cc @paulyoung


LiamGoodacre:
Thanks!

PR #153, opened 2018-03-08
bumped express to 0.6.2

spocke:
Bumped the express version the verification seems to pass. This is my first PR so hopefully I did that correctly according to the instructions.


justinwoo:
Thanks!

PR #157, opened 2018-03-15
Update the Halogen ecosystem

fosskers:
### TODO

- [x] Core `halogen` and things that depend on it
- [x] Bonus goodies like `aff` and `eff`
- [x] More `halogen-*` libs

*Note:* I didn't add `halogen-leaflet` yet, and won't in this PR.


justinwoo:
Thanks!

PR #160, opened 2018-03-23
update the universe

justinwoo:
Yup. Let's see how many need to be removed at most

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -82,7 +82,7 @@
       "integers"
     ],
     "repo": "https://github.com/purescript-contrib/purescript-argonaut-codecs.git",
-    "version": "v3.2.0"
+    "version": "v3.3.0"
   },
   "argonaut-core": {
     "dependencies": [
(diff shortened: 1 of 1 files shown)


packages.json
```diff
@@ -856,14 +856,6 @@
     "repo": "https://github.com/citizennet/purescript-halogen-select.git",
     "version": "v0.1.12"
   },
-  "halogen-storybook": {
```
justinwoo: made a PR to this repo, but the person adding this is not using this. will be removed and can be re-added later.

packages.json
```diff
@@ -2068,25 +2064,13 @@
     "repo": "https://github.com/bodil/purescript-test-unit.git",
     "version": "v13.0.0"
   },
-  "thermite": {
```
justinwoo: quite non-trivial upgrades needed here i guess. i don't know the type zoo at all in this library.

PR #162, opened 2018-03-28
Update `halogen-select` to most recent version

thomashoneyman:
The latest version of [`purescript-halogen-select`](https://github.com/citizennet/purescript-halogen-select) is a major update with some significant changes. This is definitely the version that ought to be included in psc-package. This PR updates the dependencies and version for the library.


justinwoo:
Thanks!

PR #164, opened 2018-04-09
Add Polyform

thomashoneyman:
Polyform is a composable form toolkit similar to something like `digestive-functors`. Written by @paluh -- love to see this in the default package set.


thomashoneyman:
Closed in favor of @paluh's addition.

PR #172, opened 2018-04-29
Add aff-retry package

Unisay:
https://github.com/Unisay/purescript-aff-retry
Monadic action combinators that add delayed-retry functionality, potentially with exponential-backoff, to arbitrary actions.


justinwoo:
Sorry for the delayed response, but this repo now uses a different approach for packages. Could you open a new PR if you need this package? Thanks

PR #173, opened 2018-05-01
Add webgl bindings

btrepp:
Adds the webgl package and it's dependencies.
Foreign-clone is concerning, but this is the simplest way
to add it into the package-set

I haven't forked any of the repos, preferring to add them as is.

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -555,6 +555,20 @@
     "repo": "https://github.com/nkly/purescript-express.git",
     "version": "v0.6.4"
   },
+  "extensions": {
+    "dependencies": [
+      "arrays",
+      "integers",
+      "math",
+      "lists",
+      "tailrec",
+      "foreign-clone",
+      "partial",
+      "js-date"
+    ],
+    "repo": "https://github.com/jutaro/purescript-extensions.git",
+    "version": "v2.3.8"
+  },
   "filterable": {
     "dependencies": [
       "foldable-traversable",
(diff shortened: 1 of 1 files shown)


hdgarrood:
IIRC `foreign-clone` exists because the maintainer of the webgl bindings preferred the design of an older version of the `foreign` library, and just took that older version and published it as `foreign-clone`.

btrepp:
@justinwoo I agree with the sentiment. I've only just started on the purescript journey and wanted to help contribute.

What is needed to get this curated-worthly. Just updating the deps to use foreign instead of foreign-clone?

Maybe @jutaro can comment?. This is all his good work. It would be nice to have webGL easily accessibly in the curated package set :).

AlexMouton:
@btrepp I have been working on jutaros good webgl/matrix/vector in my forks as well.
If you're digging in please take a look.  Jutaro has been unresponsive to my PRs, at least, since Dec

AlexMouton:
(I Just got my continuation of vector/matrix/webgl into a package set here: https://github.com/AlexMouton/package-sets#mouton

btrepp:
Tracing through the deps (assuming the bower.json was correct). It looks like extensions -> foreign-clone and the only place the CForiegn is used is to call 'stringify'.

@AlexMouton would you be okay maintaining a fork of webgl/matrix/vector?. I could probably give a shot at modifying 'extensions' so it doesn't depend on foreign-clone.

PR #183, opened 2018-05-27
added profunctor-lenses

CarstenKoenig:
Hi,

would be great to have this as well - although I'm not sure if I got all the dependencies right (some of those might be already trans. added by others - sorry)

also I have no clue if the version is right so that probably needs to be changed


justinwoo:
looks right, thanks

PR #191, opened 2018-06-06
add purescript-group

adkelley:
Adding [purescript-group](https://github.com/morganthomas/purescript-group)


justinwoo:
Cool, thanks!

PR #197, opened 2018-06-18
Add routing v.8.0.0, spec-quickcheck v3.0.0, record-extra v1.0.0, errors v4.0.1, web-socket v1.0.0.

Rembane:
What it says on the tin. `psc-package verify routing` worked out well.

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -279,6 +279,17 @@
     "repo": "https://github.com/purescript/purescript-enums.git",
     "version": "v4.0.0"
   },
+  "errors": {
+    "dependencies": [
+      "control",
+      "effect",
+      "either",
+      "maybe",
+      "transformers"
+    ],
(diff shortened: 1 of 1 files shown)


justinwoo:
Even as the author of record-extra I don't think I want most people to use that library. The others look fine to include.

Rembane:
@justinwoo Sweet, thank you. If I still need record-extra I'll get back to you. :)

Rembane:
I added the packages errors and web-socket too. These are all packages I need in my current project, so I won't add any more commits to this pull request, unless asked to.

Should I remove the record-extra package?

justinwoo:
Sorry for the late response, didn't realize you commented. Yes, please remove record-extra, since it's more confusing to have than not.

Rembane:
I moved the relevant parts to this pull request instead. https://github.com/purescript/package-sets/pull/200

PR #201, opened 2018-06-23
Add foreign generic package to 0.12 package set

ariaroo:
This PR just adds foreign-generic to the package set. I didn't want to disturb Phil Freeman to make this update to the psc-package set. The repo field in the entry actually points to Phil Freeman's foreign-generic github repo.

Please forgive the git diff that affected the event package. I think it was caused when I ran psc-package format.


justinwoo:
Looks good, thanks

PR #207, opened 2018-06-26
Add behaviors, bigints, formatters, remotedata, email-validate, and read

thomashoneyman:
This PR adds these repositories into the package set given their 0.12 updates:

- https://github.com/slamdata/purescript-formatters
- https://github.com/sharkdp/purescript-bigints
- https://github.com/paf31/purescript-behaviors
- https://github.com/truqu/purescript-read
- https://github.com/krisajenkins/purescript-remotedata
- https://github.com/cdepillabout/purescript-email-validate

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -148,12 +148,36 @@
     "repo": "https://github.com/slamdata/purescript-avar.git",
     "version": "v3.0.0"
   },
+  "behaviors": {
+    "dependencies": [
+      "prelude",
+      "effect",
+      "ordered-collections",
+      "filterable",
+      "nullable",
+      "event",
+      "web-html",
+      "web-events",
+      "web-uievents"
+    ],
+    "repo": "https://github.com/paf31/purescript-behaviors.git",
+    "version": "v7.0.0"
+  },
   "bifunctors": {
     "dependencies": [
       "newtype",
       "prelude"
     ],
     "repo": "https://github.com/purescript/purescript-bifunctors.git",
+    "version": "v4.0.0"
+  },
+  "bigints": {
+    "dependencies": [
+      "maybe",
+      "integers",
+      "strings",
+    ],
+    "repo": "https://github.com/sharkdp/purescript-bigints.git",
     "version": "v4.0.0"
   },
   "canvas": {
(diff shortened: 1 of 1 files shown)


thomashoneyman:
Well, it turns out those packages don't build together for various reasons (among them, other missing packages from the ecosystem). I've updated with new packages, which now includes:

- https://github.com/paf31/purescript-behaviors
- https://github.com/sharkdp/purescript-bigints
- https://github.com/paf31/purescript-behaviors (updated)
- https://github.com/slamdata/purescript-fixed-points
- https://github.com/slamdata/purescript-formatters
- https://github.com/citizennet/purescript-fuzzy
- https://github.com/citizennet/purescript-halogen-select
- https://github.com/purescript-contrib/purescript-parsing
- https://github.com/paluh/purescript-polyform
- https://github.com/anttih/purescript-rationals
- https://github.com/truqu/purescript-read
- https://github.com/krisajenkins/purescript-remotedata
- https://github.com/natefaubion/purescript-run
- https://github.com/purescript-contrib/purescript-strongcheck

I'd consider all of these packages worthy of inclusion in the curated package set.

PS. It's pretty tedious putting these together and verifying the build, and packages that are viable to be included in new sets have to be done manually or else get dropped out (like many of the above). Any ideas in the works to make this faster & preserve libraries between sets?

thomashoneyman:
Also @justinwoo for some reason `email-validate` installs just fine with Bower, but refuses to verify with the package set with some missing `generics-rep` dependencies. Any idea why that is?

thomashoneyman:
Well, now it's OK. Not sure what was going on. So we'll include that one too:

- https://github.com/cdepillabout/purescript-email-validate

justinwoo:
How were you adding these dependencies? Usually if I'm to add them from bower, I use `psc-package add-from-bower` and verify the packages I added each by doing `psc-package verify package-name`. Make sure you clear `output` and `.psc-package` if you run into inconsistencies between CI and your local.

> curated package set

I wouldn't call this a "curated" package set since there's not enough of a notion of that, it's just whatever community sourced packages for things that are actually generally useful.

> Any ideas in the works to make this faster 
I don't have anything that I'm putting in this repo. I have my package sets with Dhall (https://github.com/justinwoo/spacchetti) and I work by using some Dhall codegen and manual editing, and it takes me a minute or two to add a new package and verify it locally.

> & preserve libraries between sets?

No real ideas on that front. Every user should own their own package sets and not depend on this repo though. This package set is more of a template that users can work off of or scrape information from.

justinwoo:
I'll merge this since it looks fine, thanks

thomashoneyman:
Thanks for the info.

> This package set is more of a template that users can work off of or scrape information from.

I'd actually been thinking of this a bit like Stackage snapshots -- an attempt to collect as much as the ecosystem as possible on their latest compatible versions. But if this is meant to be a way to simply list your project dependencies against a small personal snapshot, then perhaps there's an opportunity to update documentation here and provide guidance on that. For now we'll stick with this big package set and diverge as necessary!

PR #223, opened 2018-08-27
Add Typelevel and Hedwig

danieljharvey:
Hi!

These have come in useful for me recently so thought I would share.

Thanks!
Dan


justinwoo:
Thanks

PR #245, opened 2018-11-10
react-basic: v3.0.0 -> v5.0.0

mb21:
As you can see from the version numbers, it's a breaking change.

See https://github.com/lumihq/purescript-react-basic/#migrating-from-v2-or-v3


justinwoo:
Thanks!

PR #248, opened 2018-11-28
Update spec to v3.1.0

felixmulder:
This PR updates the spec libraries to v3.1.0 and removes a dev dependency from presto.


justinwoo:
Thanks

PR #264, opened 2019-02-09
Add purescript-leibniz to package.json

DonaldKellett:
A package with a single module `Data.Leibniz` for expressing type equality.

Pursuit documentation: https://pursuit.purescript.org/packages/purescript-leibniz/5.0.0


justinwoo:
Thanks

PR #277, opened 2019-03-02
Update docs

### LINKED ISSUE #128: Tag automatically
CI checks each PR that the set still works. But tagging is still a manual process and lags behind changes. Can we make it so every merge to master creates a tag?

### LINKED ISSUE #276: Write package author guide
TODO

f-f:
This is a general adjustment of documentation after the Spacchetti merge. This PR includes:
- rewrite of the `README` to detail only how to use the package-set
- addition of `CONTRIBUTING.md` to detail some contribution policies and a detailed example of how to add packages to the set (note that the examples are now in ascii)
- addition of table of contents for both the readme and the contribution guide

Perhaps the most important part of this PR is the definition of some guidelines for the package-set, e.g.:
- a policy for if/when packages are dropped from the set
- a set of criterias new packages should comply, based on [this thread](https://twitter.com/jusrin00/status/1100655242264412160)

Fix #128 
Fix #276


CONTRIBUTING.md
```diff
@@ -0,0 +1,234 @@
...
+Note: if the package you're adding has a [spago][spago] configuration, you can
+simply import the dependencies instead of typing them down. E.g. the above example would
+look like:
+
+```hs
+    let mkPackage = ./../mkPackage.dhall
+
+in  { unicorns =
+        mkPackage
+        (https://raw.githubusercontent.com/someauthor/unicorns/v4.2.0/spago.dhall).dependencies
```
f-f: @justinwoo I added here an example of how to add spago-libraries. I don't think this will be common (since ~100% of libraries so far are bower-based), but I'd like to try this mode out - I can remove all the remote imports if they ever become a problem

justinwoo: i will end up removing these when i update them even if they do

f-f: Which are the reasons that make you say this?

justinwoo: it's what my script does when running

README.md
```diff
@@ -1,33 +1,59 @@
-# Package-Sets
+# package-sets
```
justinwoo: why

README.md
```diff
@@ -1,33 +1,59 @@
...
 
 [![Build Status](https://travis-ci.org/purescript/package-sets.svg?branch=master)](https://travis-ci.org/purescript/package-sets)
 
-PureScript package sets for Psc-Package, aka "Spacchetti"
+A curated list of PureScript packages for Psc-Package and Spago
```
justinwoo: you might add "and Spago" but the other changes should be reverted

README.md
```diff
@@ -1,33 +1,59 @@
...
 
 [![Build Status](https://travis-ci.org/purescript/package-sets.svg?branch=master)](https://travis-ci.org/purescript/package-sets)
 
-PureScript package sets for Psc-Package, aka "Spacchetti"
+A curated list of PureScript packages for Psc-Package and Spago 
 
 ![](https://i.imgur.com/roCuNQ9.png)
 
-Dhall-driven package sets, made for forking and modifying easily.
+*Historical note*: this project used to be called [spacchetti][spacchetti]
```
justinwoo: this note is also here already

CONTRIBUTING.md
f-f: I assume you're talking about [this script](https://github.com/justinwoo/spac-update/blob/master/src/main.rs). I can update it when we get the first case of this

README.md
```diff
@@ -1,33 +1,59 @@
...
+- [Add your package](#add-your-package)
+- [How to use package-sets?](#how-to-use-package-sets)
+  - [..with `psc-package`](#with-psc-package)
+  - [..with `spago`](#with-spago)
 
-[Psc-Package](https://github.com/purescript/psc-package/) is a package manager for PureScript that works essentially by running a bunch of git commands. Its distinguishing feature from most package managers is that it uses a **package set**.
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
 
-### What is a Package Set?
+## What is a package-set?
```
justinwoo: should be "Package Set", not hyphenated

README.md
f-f: Because it was like that [before the spacchetti merge](https://github.com/purescript/package-sets/blob/0681dece0535270425f2a376e757e4484b1b4c52/README.md)

README.md
f-f: Do we have a naming convention somewhere? I've been using the hyphenated version all my life (e.g. all over this PR, and in Spago docs)

README.md
```diff
@@ -1,33 +1,59 @@
...
 
-Many users trying to rush into using Psc-Package don't slow down enough to learn what package sets are. They are a **set** of packages, such that the package set only contains **one** entry for a given package in a set. This means that
+A package set is a **collection** of packages, such that there is only **one** entry
+(i.e. version) for every given package in the set.
```
justinwoo: Could you remove the column wrapping?

README.md
f-f: I'm not sure what do you mean by this?

README.md
```diff
@@ -1,33 +1,59 @@
...
-### How are package sets used?
+This repository aims to be a good collection of packages you can depend on.  
+In general we welcome all packages, provided that they follow some guidelines
+defined in the [contributing guide](#CONTRIBUTING.md).
 
-Package sets are consumed by having a `psc-package.json` file in the root of your project, where the contents are like below:
+The linked document also contains instructions on how to add new packages to the
+set, and information on versioning and related policies.
+
+## How to use package-sets?
```
justinwoo: "How are package sets used" or "How do I use package sets?"

README.md
f-f: Only for this line?

README.md
```diff
@@ -1,33 +1,59 @@
...
+
+### ..with `psc-package`
```
justinwoo: just "Usage with Psc-Package" if you insist

README.md
```diff
@@ -42,178 +68,57 @@ Package sets are consumed by having a `psc-package.json` file in the root of you
...
 
-See the Spago repo for more: <https://github.com/spacchetti/spago>
+You can change the package-set version you are using by just pointing the `upstream` to
+a different location, and running `spago freeze` afterwards (see [its readme][spago]
+for more info about this)
 
-### Does Spago replace Psc-Package?
 
-Yes and no. See the Spago project for the full details on what it is and how it is the similar and different from Psc-Package: <https://github.com/spacchetti/spago>
+[spacchetti]: https://github.com/spacchetti/spacchetti
```
justinwoo: do we need this? i.e. have we not already started rewriting everything from this?

CONTRIBUTING.md
```diff
@@ -0,0 +1,234 @@
+# Contributing to package-sets
```
justinwoo: Package Sets or package sets

README.md
f-f: Here I just matched the description of the repo that we had before the merge.

The reason for this wording is that the assumption in the readme is that the reader doesn't know what a package set is (it's literally the first thing we explain), so "curated list of packages" is a good preliminary definition

CONTRIBUTING.md
```diff
@@ -0,0 +1,234 @@
...
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+## General
+
+All changes go through pull requests.
+
+There are some criterias that packages being added should comply to:
+- If you use [bower][bower], you should not include your `devDependencies`
+  in your `dependencies`.
```
justinwoo: This part doesn't really matter, but the following one does

CONTRIBUTING.md
```diff
@@ -0,0 +1,234 @@
...
+## General
+
+All changes go through pull requests.
+
+There are some criterias that packages being added should comply to:
+- If you use [bower][bower], you should not include your `devDependencies`
+  in your `dependencies`.  
+  You can test if your bower package works by running `bower i -p`
+- The published tag should start with the `v` prefix.  
+  If you use [pulp][pulp] to version and publish your packages, everything should be fine.
```
justinwoo: would also be easier to just tell people to actually use `pulp version` and `pulp publish` commands

CONTRIBUTING.md
```diff
@@ -0,0 +1,234 @@
...
+
+## Releases
+
+A "release" of the package set consists in a *git tag* (i.e. what GitHub calls a release).
+
+Releases happen quite often (look at the [release history][releases]), and anyone can request
+a new release at any time, by just [opening an issue][issues].  
+This implies that the `master` branch should always be "release ready".
+
+If you use the Dhall version of the set (for either [`psc-package`][local-setup] or `spago`),
```
justinwoo: I don't want to use this for package sets. I will manually edit these out when they come in otherwise.

README.md
f-f: I worded it this way so that it would match the `spago` section. Should I rename that too?

CONTRIBUTING.md
justinwoo: If this note is about using the package sets in general, then it should go in the Spago README, not contributors' notes in package sets.

README.md
justinwoo: No, in general

CONTRIBUTING.md
```diff
@@ -0,0 +1,234 @@
+# Contributing to package-sets
+
+This document is meant to be a general guide and FAQ for contributors to this repo.
+
+It defines some policies that are applied, and details how to add new packages and the
+criteria packages should match to stay in the set.
+
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
```
justinwoo: I don't really want to use doctoc personally

README.md
f-f: Ok

README.md
f-f: Alright

CONTRIBUTING.md
f-f: Cool, I'll take it away

CONTRIBUTING.md
f-f: Sounds good

CONTRIBUTING.md
f-f: Alright, I'll move it to Spago README. The reason why I put it in was to resolve #128, which was then psc-package-specific by this criteria

CONTRIBUTING.md
f-f: Do you have some nicer way that doesn't involve doing this work by hand?

CONTRIBUTING.md
justinwoo: this should have notes that package authors should publish their packages on bower and ensure `bower i -p` works.

justinwoo: this might change in 2020, but for 2019 it will likely not change

README.md
justinwoo: thx

CONTRIBUTING.md
f-f: `bower i -p` is in one of the notes you suggested to take away though

README.md
justinwoo: would be good

README.md
justinwoo: this PR removes the same thing being already written

README.md
justinwoo: it's not really curated though, that's just some historical misnaming

CONTRIBUTING.md
justinwoo: as i said earlier, i also want all of the static data to be here, rather than to have remote imports in random places.

README.md
f-f: Oh. I thought this historical note was referring to other materials outside of this repo. If it refers to the repo itself I can take it away

CONTRIBUTING.md
justinwoo: also related that i support packages that are actually published via pulp, and it works if they have even the most minimal setup using "*" versions

CONTRIBUTING.md
justinwoo: readthedocs :+1:

CONTRIBUTING.md
f-f: This refers to the name of the repo

README.md
f-f: Ok

README.md
f-f: I'd argue that there's a good amount of curation going on :)

CONTRIBUTING.md
f-f: That would empty the readme once again, which I don't think it's desiderable.

I can take away the ToC, or just the doctoc annotations if you wish

justinwoo: yeah i'd like to just remove the TOC or only manually have some important entries

justinwoo:
We'll also see what all to remove or add later over time

f-f:
@justinwoo incorporated feedback

README.md
justinwoo: There really isn't much at all, I really would not say this ever

CONTRIBUTING.md
justinwoo: `package-sets`

README.md
justinwoo:
```suggestion
# Package Sets
```

README.md
```diff
@@ -1,33 +1,40 @@
...
-Historically, this was a separate project created by Justin Woo called "Spacchetti" (as in Spagghetti + Pacchetti). This has now merged into package-sets, but is still referred to as "Spacchetti" in many resources.
+- [What is a package-set?](#what-is-a-package-set)
+- [Add your package](#add-your-package)
+- [How do I use package-sets with `psc-package`?](#how-do-i-use-package-sets-with-psc-package)
+- [How do I use package-sets with `spago`?](#how-do-i-use-package-sets-with-spago)
 
-## Introduction to Psc-Package (from a Package-Sets perspective)
 
-### What is Psc-Package?
+## What is a package-set?
```
justinwoo:
```suggestion
## What is a package set?
```

README.md
```diff
@@ -42,178 +49,49 @@ Package sets are consumed by having a `psc-package.json` file in the root of you
...
-### Testing changes to package set
-
-To set up a test project, run `make setup`. Then you can test individual packages with `psc-package verify PACKAGE`.
-
-### Package set maintenance
-
-If you would like to help maintain Package-Sets, please get in touch with Justin via Twitter: <https://twitter.com/jusrin00>
-
-## Spago
+## How do I use package-sets with `spago`?
```
justinwoo:
```suggestion
## How do I use `package-sets` with `spago`?
```

README.md
```diff
@@ -1,33 +1,40 @@
...
-* Whichever package you want to install must be in the package set
-* The dependencies and the transitive dependencies of the package you want to install must be in the package set
+This repository aims to be a good collection of packages you can depend on.  
+In general we welcome all packages, provided that they follow some guidelines defined in the [contributing guide](#CONTRIBUTING.md).
 
-Package sets are defined in `packages.json` in the root of any package set repository, like in <https://github.com/purescript/package-sets/blob/master/packages.json>.
+The linked document also contains instructions on how to add new packages to the set, and information on versioning and related policies.
 
-### How are package sets used?
+## How do I use package-sets with `psc-package`?
```
justinwoo:
```suggestion
## How do I use `package-sets` with Psc Package?
```

README.md
```diff
@@ -1,33 +1,40 @@
...
+- [How do I use package-sets with `psc-package`?](#how-do-i-use-package-sets-with-psc-package)
+- [How do I use package-sets with `spago`?](#how-do-i-use-package-sets-with-spago)
 
-## Introduction to Psc-Package (from a Package-Sets perspective)
 
-### What is Psc-Package?
+## What is a package-set?
 
-[Psc-Package](https://github.com/purescript/psc-package/) is a package manager for PureScript that works essentially by running a bunch of git commands. Its distinguishing feature from most package managers is that it uses a **package set**.
+A package set is a **collection** of packages, such that there is only **one** entry (i.e. version) for every given package in the set.
```
justinwoo:
```suggestion
A package set is a **collection** of packages, such that there is only **one** entry (i.e. version) for a given package in the set.
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,190 @@
...
+All changes go through pull requests.
+
+There are some criterias that packages being added should comply to.
+
+If you use [bower][bower]:
+- `bower i -p` should run succesfully
+- The published tag should start with the `v` prefix.  
+  In general you should use `pulp version` and `pulp publish`, and everything will be fine.
+
+If you don't use bower, get in touch by [opening an issue][issues], as the maintainers' tools might need some adjustments.
```
justinwoo: No, they should actually use Pulp to publish packages

README.md
f-f: The assumption of the reader not having to know what a package set is still stands, so I'd like to see some alternative proposals here

CONTRIBUTING.md
hdgarrood: I'm not sure if this has always been the case for this package set or if there was a different policy in spacchetti previously or whatever, but I'd really strongly prefer that packages be required to be registered on bower before they may be included in this package set, and that they have the same name in the package set that they do on bower (minus the "purescript" prefix). Pursuit deliberately only accepts packages which are on bower at the moment, because bower gives us a central registry for names so that there can never be disagreement about what a given package name refers to.

f-f: Thanks @hdgarrood, your comment is very helpful and I appreciate that you took the time to write it :)

The context here is that since this is a documentation effort, in order to give contributors actually meaningful guidelines I'm trying to find out:
1. what are the actual requirements for packages to get to the package set
2. why are they in place (this includes technical and historical reasons)

..so that we can write down this information in the above paragraph.

It has always been the case in spacchetti that all packages were published on bower, but to my knowledge this was not documented anywhere (or if this documentation exists no one has pointed me at it yet)

I'll try to recap the reasons in your comment:
- "the package must be on bower" is an actual requirement
- and the reason for that is "Pursuit only accepts packages that are on bower"
- and the reason for that is "bower provides a central registry for names"

Now, a package set is also a "central registry for names" (so it would solve the same problem as bower), but we are picking bower as the single source of truth because then there's no possibility of divergence between the "bower ecosystem" and the "package set ecosystem".

@hdgarrood does this sound accurate?

hdgarrood: Yes, I think that's a good way of putting it. :+1:

Although actually on second thoughts, what I wrote originally is not quite true: it is possible to publish packages to Pursuit which are not on the Bower registry, since Pursuit doesn't try to confirm that you have published your package to Bower. As far as I'm aware, though, nobody has managed to upload packages to Pursuit which are not on Bower, because to do that you'd basically have to reimplement or manually perform most of the steps that `pulp publish` takes care of for you, as well as putting together a manifest file which parses as a valid bower.json file. So the maybe the difference is academic (in the sense of not actually mattering in the real world).

f-f: @hdgarrood @justinwoo I updated the above section with the results from this thread (in case you're wondering, I left out any mention of Pursuit since it's only a transitive step in the cause analysis)

CONTRIBUTING.md
```diff
@@ -0,0 +1,188 @@
...
+    - [Prerequisites](#prerequisites)
+  - [1. Adding a new package](#1-adding-a-new-package)
+  - [2. Verifying a package](#2-verifying-a-package)
+
+
+## General
+
+All changes go through pull requests.
+
+There are some criterias that packages being added should comply to:
```
hdgarrood: Some grammar nits:
- "Criteria" is already plural, so there's no need for the extra 's'. (In case you're wondering, the singular is "criterion".)
- It's "comply with" rather than "comply to".

CONTRIBUTING.md
hdgarrood: Looks great, thanks! :+1: from me.

CONTRIBUTING.md
f-f: Thanks a lot for this! I did study Latin for too many years, but English has a weird treatment of latin words anyways so I never know if I'm supposed to latin-conjugate them or not 😅 (my pet peeve here is "caveat"..)

justinwoo:
needs rebase then let's merge it

PR #285, opened 2019-03-22
Add monad-logger v1.0.0

cprussin:
Add monad-logger v1.0.0.

Also update the `default.nix` file to include some missing dependencies (dhall and git).

Would it be too much to request a release after this is merged in?

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -14,5 +14,8 @@ in pkgs.runCommand "easy-ps-test" {
     purs
     psc-package
     dhall-json-simple;
-  };
+  } ++ [
+    pkgs.dhall
+    pkgs.git
+  ];
 } ""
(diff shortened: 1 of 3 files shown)


cprussin:
FYI @justinwoo

default.nix
```diff
@@ -14,5 +14,8 @@ in pkgs.runCommand "easy-ps-test" {
     purs
     psc-package
     dhall-json-simple;
-  };
+  } ++ [
+    pkgs.dhall
```
justinwoo: Should use dhall from easy-ps, just add dhall to the list above. Otherwise, I cannot use this.

cprussin: fixed, thanks

justinwoo:
Thanks

PR #288, opened 2019-03-29
Clarify requirements for packages being added to the set

JordanMartinez:
@f-f As promised in [my comment](https://github.com/spacchetti/spago/issues/142#issuecomment-475465232)

Is there anything else we want to change here to make things clearer? For example, while `pulp version` may handle the "v" prefix, are there any other requirements (e.g. use semver for versioning)?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -22,9 +22,7 @@ It defines some policies that are applied, and details how to add new packages a
 
 All changes go through pull requests.
 
-All packages that are included here must first be published via `bower` with no exceptions. We want to avoid a similar situation that happened in Haskell: a split ecosystem. (If you're unfamiliar with this situation, read [Why not both](https://medium.com/@fommil/why-not-both-8adadb71a5ed) for some context. Their `cabal` is similar to our `bower` and their `stack` is similar to `spago`/`psc-package`.)
-
-Since there are two distribution methods for packages (the Bower registry and the package sets), we rely on the Bower registry to act as a "central registry of package names" for both methods. This prevents divergence in the ecosystem - e.g. having two different codebases for a package called "prelude".
+All packages that are included here must first be published via `bower` with no exceptions. Since there are two distribution methods for packages (the Bower registry and the package sets), we rely on the Bower registry to act as a "central registry of package names" for both methods. This prevents divergence in the ecosystem - e.g. having two different codebases for a package called "prelude".
 
 Packages must comply with the following criteria. The `pulp` commands listed below are used because they handle most of the work for you:
 - `bower i -p` must run successfully.  
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -24,7 +24,7 @@ All changes go through pull requests.
 
 All packages that are included here must first be published via `bower` with no exceptions. We want to avoid a similar situation that happened in Haskell: a split ecosystem. (If you're unfamiliar with this situation, read [Why not both](https://medium.com/@fommil/why-not-both-8adadb71a5ed) for some context. Their `cabal` is similar to our `bower` and their `stack` is similar to `spago`/`psc-package`.)
 
-Since there are two distribution methods for packages (the Bower registry and the package sets), we rely on the Bower registry to act as a "central registry of package names" for both methods. This prevents divergence in the ecosystem similar to the Haskell situation described above (e.g. having two different codebases for a package called "prelude").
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -22,12 +22,17 @@ It defines some policies that are applied, and details how to add new packages a
 
 All changes go through pull requests.
 
-There are some criteria that packages being added should comply with:
-- _You should use `pulp version` to version your package._  
-  The reason is that the tag should start with the `v` prefix
-- _You should use `pulp publish` to publish your package._  
-  The reason is that since there are two distribution methods for packages (the Bower registry and the package sets), we rely on the Bower registry to act as a "central registry of package names" for both methods, in order to prevent divergence in the ecosystem (e.g. having two different codebases for a package called "prelude")
-- `bower i -p` should run successfully
+All packages that are included here must first be published via `bower` with no exceptions. We want to avoid a similar situation that happened in Haskell: a split ecosystem. (If you're unfamiliar with this situation, read [Why not both](https://medium.com/@fommil/why-not-both-8adadb71a5ed) for some context. Their `cabal` is similar to our `bower` and their `stack` is similar to `spago`/`psc-package`.)
```
f-f:
```suggestion
All packages that are included here must first be published via `bower` with no exceptions.
```

AFAIK there hasn't been any split in the Haskell ecosystem: all the packages on Stackage _have to_ be published on Hackage. They use Hackage basically in the same way we use the Bower registry.

Also I'm not sure I'd like to mention the whole Haskell packages situation: it's quite controversial there, but in here we're totally fine with it. So this risks giving the wrong impression to the reader.

It might be just me though, so I'd like to hear @justinwoo's opinion on this too

CONTRIBUTING.md
```diff
@@ -22,12 +22,17 @@ It defines some policies that are applied, and details how to add new packages a
...
+
+Since there are two distribution methods for packages (the Bower registry and the package sets), we rely on the Bower registry to act as a "central registry of package names" for both methods. This prevents divergence in the ecosystem similar to the Haskell situation described above (e.g. having two different codebases for a package called "prelude").
```
f-f:
```suggestion
Since there are two distribution methods for packages (the Bower registry and the package sets), we rely on the Bower registry to act as a "central registry of package names" for both methods. This prevents divergence in the ecosystem - e.g. having two different codebases for a package called "prelude".
```

f-f:
@JordanMartinez there's no requirement on using semver for versioning, since the package upgrades have to be tested anyways: semver is made by humans, but the compiler has anyways the last word on the question "does the package set compile?", so we cannot trust it

CONTRIBUTING.md
JordanMartinez: > AFAIK there hasn't been any split in the Haskell ecosystem: all the packages on Stackage have to be published on Hackage. They use Hackage basically in the same way we use the Bower registry.

Oh. I'm not familiar with how Haskell's build tools work but I know this was one of those situations that affects many in that ecosystem.

Ok. Once Justin gives his feedback, we'll continue from there. If I need to delete that section, that's fine.

CONTRIBUTING.md
JordanMartinez: Done.

JordanMartinez:
> there's no requirement on using semver for versioning, since the package upgrades have to be tested anyways: semver is made by humans, but the compiler has anyways the last word on the question "does the package set compile?", so we cannot trust semver

@f-f Thanks for clarifying. Should I explain that in the guidelines or leave it as is? While the package sets might not care about this, `bower` does. As long as `bower` is still used as a central registry, then it matters somewhat.

f-f:
@JordanMartinez 
> Should I explain that in the guidelines or leave it as is? While the package sets might not care about this, bower does. As long as bower is still used as a central registry, then it matters somewhat.

It does matter for bower, but it's not a requirement to get into the package-set, which is what we're detailing here. So I'd say we should not mention it in the guidelines

JordanMartinez:
Sounds good.

CONTRIBUTING.md
justinwoo: Yeah, I think the only relation here is kind of like how all Haskell packages are on Hackage and then consumed downstream in Stackage. Maybe there might be some kind of useful comparison in the future, but I think it would be better to just leave this out, lest we be charged with inciting holy wars in Haskell land

JordanMartinez: Ok. I've deleted that section in the latest commit.

f-f:
Thanks @JordanMartinez!

PR #289, opened 2019-03-29
Rev monad-logger to v1.1.0

cprussin:
@justinwoo I'm not sure how much of an ask it is to get you to do releases.  If it's easy enough I'd love to get it released, otherwise I can just leave the override in my spago config until the next release happens.  Thanks!


justinwoo:
Generally we try to release after each merge since it's by date anyway. Thanks

PR #292, opened 2019-04-04
Re-add node-electron

cprussin:
@justinwoo I somehow didn't ever see your [request to publish this package to bower](https://github.com/cprussin/purescript-node-electron/issues/1) and I just noticed that the package got dropped from the package set.  My apologies on missing that ticket!

I just revved the package and published to bower (I did not use `pulp version` since I'm running into the OOM issue that I understand is being worked on in pulp--let me know if that's a problem), I'd love to get this re-added to the package set now.  Thanks a bunch!

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -1442,6 +1442,23 @@
     "repo": "https://github.com/purescript-node/purescript-node-child-process.git",
     "version": "v6.0.0"
   },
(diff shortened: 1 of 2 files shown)


justinwoo:
thanks

PR #296, opened 2019-04-10
 changed `/bin/bash` to `/usr/bin/env bash`

8084:
Can we have this little change to ensure compatibility with systems that use non-standard paths, e.g. NixOS?


justinwoo:
we can try it

PR #307, opened 2019-05-08
Add package "result" by @ad-si

ad-si:


f-f:
@ad-si you should run `make` to generate the new `packages.json` so CI will pass

f-f:
Thanks!

PR #308, opened 2019-05-09
Add missing `make` step

ad-si:
I followed the guide, and this was apparently missing => https://github.com/purescript/package-sets/pull/307

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -70,8 +70,8 @@ The following section will detail how to add a package to the package-set.
 
 The *TL;DR* about it is:
 - add the Dhall package definition in some `src/groups/${username}.dhall` (where `username` is the one of the author of the package)
+- add a new line containing `⫽ ./groups/${username}.dhall` to `src/packages.dhall`
 - run `make setup` and `psc-package verify ${your-new-package-name}`
-- run `make` to add an entry for your package to `packages.json`
 
 ### 0. Background knowledge
 
@@ -168,6 +168,16 @@ in  { unicorns =
     }
 ```
 
+Then add a new line containing a reference to the new group to the `src/packages.dhall` file.
+
+For our example: 
+
+```hs
+…
+      ⫽ ./groups/someauthor.dhall
+…
+``` 
+
 
 ### 2. Verifying a package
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -71,6 +71,7 @@ The following section will detail how to add a package to the package-set.
 The *TL;DR* about it is:
 - add the Dhall package definition in some `src/groups/${username}.dhall` (where `username` is the one of the author of the package)
 - run `make setup` and `psc-package verify ${your-new-package-name}`
+- run `make` to add an entry for your package to `packages.json`
```
f-f: Running the line above should already do this though? 🤔 (see [the makefile](https://github.com/purescript/package-sets/blob/180014f95bdad5ca6a84959664cd05f27c350566/Makefile#L13))

CONTRIBUTING.md
```diff
@@ -176,7 +177,8 @@ In order to verify the addition (or change), you should follow these steps:
 - `make setup`: this will setup a test project based on the new package-set
 - `psc-package verify ${your-new-package-name}`
 
-Once it verifies correctly, you should check in both the Dhall files and the `packages.json`
+Once it verifies correctly, you should run `make` to add an entry for your package to `packages.json`
+and finally check in both the Dhall files and the `packages.json` file.
```
f-f: Same as above about the `make setup`, but let's keep this line about checking in both files!

CONTRIBUTING.md
ad-si: Ahh I think the problem was that you have to create the `src/groups/${username}.dhall` file *and* add an entry to `src/packages.dhall`, but that's not mentioned in the guide. Then `make setup` is enough

CONTRIBUTING.md
```diff
@@ -70,6 +70,7 @@ The following section will detail how to add a package to the package-set.
 
 The *TL;DR* about it is:
 - add the Dhall package definition in some `src/groups/${username}.dhall` (where `username` is the one of the author of the package)
+- add a new line containing `⫽ ./groups/${username}.dhall` to `src/packages.dhall`
```
f-f:
```suggestion
- if adding a new group file, also add a new line containing `⫽ ./groups/${username}.dhall` to `src/packages.dhall`
```

f-f:
Thanks!

PR #327, opened 2019-06-01
Update prelude, partial

spacchettibotti:
The addition has been verified, this is safe to merge.


justinwoo:
"verified"

f-f:
This means that "spago verify-set" has been run locally, so it's verified until we come up with a better/different criteria

PR #343, opened 2019-06-06
Add purescript-metadata package to track the compiler version

### LINKED ISSUE #96: Add the supported compiler versions to the package set JSON
See https://github.com/purescript/psc-package/issues/47 for context.

f-f:
Fix #96 by adding a bogus package having the same SemVer as the compiler, so that tools can make use of this information.

If we'll need more metadata about the compiler and/or the package set we can always add it _inside_ this package.

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -1390,6 +1390,11 @@
     "repo": "https://github.com/paf31/purescript-memoize.git",
     "version": "v5.0.0"
   },
+  "metadata": {
+    "dependencies": [],
+    "repo": "https://github.com/spacchetti/purescript-metadata.git",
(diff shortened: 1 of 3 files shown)


justinwoo:
metadata project needs to have a source file

f-f:
Yep I just realized that. I'll `push -f` to that tag 😱

justinwoo:
you migth need to delete the travis cache in this branch

f-f:
Right, cool thanks

f-f:
@justinwoo this should be ready now 🎉

justinwoo:
ship it

PR #347, opened 2019-06-09
Update spec-discovery to v4.0.0

drewolson:
`v4.0.0` fixes an issue with spec discovery on the `4.0.0-rc1` version of `spec`.


justinwoo:
thanks

PR #354, opened 2019-06-10
To lowercase

oreshinya:


justinwoo:
it should be fine to leave these as is, only the filenames need to be lower cased (because of OSX users)

f-f:
Yes, and also this is a bug in the bot, so that's the only thing that should be patched

oreshinya:
Because I felt sorry to make you feel that you have to fix it soon, I did it for the current curator's problem.
Sorry, I close it.

f-f:
@oreshinya no worries, I already fixed the bug in the Curator so the next updates should be automated. And I appreciated, thanks 🙂

PR #365, opened 2019-06-22
add node-net

bouzuya:
https://github.com/purescript-node/purescript-node-net/blob/v1.0.0/bower.json


justinwoo:
Needs to run `make`, but I think it's fine otherwise

justinwoo:
Thanks

PR #375, opened 2019-06-26
generate packages.dhall

justinwoo:
This is nice


justinwoo:
"why don't we just edit one giant dhall file, just like old times?"

-> now we're cooking with gas

Makefile
```diff
@@ -7,6 +7,7 @@ format:
 generate: SHELL:=/usr/bin/env bash
 generate:
 	@dhall-to-json --pretty <<< "./src/packages.dhall" > packages.json
+	@dhall <<< ./src/packages.dhall > packages.dhall
```
f-f: Soon we might need to evolve the schema of this file, can we call this one `packagesV1.dhall`? (so the next ones will follow a coherent naming scheme)

justinwoo: why not go the whole hog and call it v1.dhall? the URL itself is already package-sets

justinwoo: nvm, i forgot about hte gigantic hash that prepends the filename, so it's a little confusing

f-f: I'd also prefer `v1.dhall`, I don't think it would be confusing

f-f:
@justinwoo I'll close this for now so I don't accidentally merge it, let's reopen in a couple of releases if we still feel this is slow

justinwoo:
I think I will do this sooner than later though, since it is really slow for me. I'll make a new one later

justinwoo:
Maybe one helpful patch is just to make Spago generate a packages.dhall once, and I can then copy that around as I wish

f-f:
@justinwoo we cannot really see the effect in less than 2-4 releases (technically 2 are enough but you want more to account for variance in amount of changes)

Something that would help would be to import single packages inside the `packages.dhall`, instead of importing single groups (we can organize the file so that we can still merge together "groups", so you can comment out much stuff at the same time), I can put together a PR for this

PR #377, opened 2019-06-28
Update dotenv to v0.4.1

spacchettibotti:
The addition has been verified by running `spago verify-set` in a clean project, so this is safe to merge.

Link to release: https://github.com/nsaunders/purescript-dotenv/releases/tag/v0.4.1

PR #378, opened 2019-06-29
Update dotenv to v0.4.1 [dependencies added]

nsaunders:
I'm expecting CI to fail (see #377), but these are the correct changes for dotenv v0.4.1. #377 upgraded the dotenv version, but it did not include the new dependencies (run and sunde).

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -615,10 +615,12 @@
       "parsing",
       "prelude",
       "psci-support",
-      "spec"
+      "run",
+      "spec",
+      "sunde"
     ],
     "repo": "https://github.com/nsaunders/purescript-dotenv.git",
-    "version": "v0.3.0"
+    "version": "v0.4.1"
   },
   "effect": {
     "dependencies": [
--- a/src/groups/nsaunders.dhall
+++ b/src/groups/nsaunders.dhall
@@ -8,8 +8,10 @@ in  { dotenv =
         , "parsing"
         , "prelude"
         , "psci-support"
+        , "run"
         , "spec"
+        , "sunde"
         ]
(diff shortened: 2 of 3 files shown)


justinwoo:
Did you run `make` on this?

nsaunders:
Fair question. In fact, `make setup` reverts packages.json to the previous version--no good! However, I found two ways to resolve (or work around?) this. I have two corresponding PRs in my fork targeting this branch; so please tell me which one to merge, and it should fix the CI check.

1. [The first one](https://github.com/nsaunders/package-sets/pull/7) (clearly the less desirable option, IMO) reverses the changes that were introduced in #374.
2. [The second one](https://github.com/nsaunders/package-sets/pull/8) updates the hash for the nsaunders group (containing dotenv). I did this by removing the hash from src/packages.dhall and running `make setup` again.

I assume the second solution is the correct one, but I suppose maybe the contributor docs need a small update to mention this extra step.

Thoughts?

justinwoo:
@f-f why is this busted?

justinwoo:
apparently Dhall doesn't do its job, so we have to do this: https://github.com/purescript/package-sets/pull/379

justinwoo:
The reversion is a result of the cache entry for the previous hash being used, so removing ~/.cache/dhall will make it actually update the hash too.

nsaunders:
Awesome, thanks for the solution @justinwoo!

justinwoo:
:+1:

f-f:
Sorry for the delay, happy you figured 🙂 

That's indeed the correct solution, and I wonder if it'd be worth to add a flag to `dhall freeze` so that it would have this behaviour 🤔

PR #399, opened 2019-07-12
remove mkPackage and define records directly

justinwoo:
just not very useful when it doesn't construct a record of newtyped values.


justinwoo:
this does not yet delete the actual mkPackage function

justinwoo:
we'll see in next version if homogeneous map is actually useful

CONTRIBUTING.md
```diff
@@ -148,24 +135,22 @@ In order to hack on this project, you should have installed:
 
 ### 1. Adding a new package
 
-To add a new package to the package set, you should create a package definition with `mkPackage`, and put it in the group file corresponding to the author's username.
+To add a new package to the package set, you should create a package definition matching the Package type, and put it in the group file corresponding to the author's username.
 
 For example, if I wish to add to the package-set the version `v4.2.0` of the package `unicorns` from `someauthor`, I will create the file `src/groups/someauthor.dhall`.
```
justinwoo:
```suggestion
For example, if I wish to add to the package-set the version `v5.0.0` of the package `some-food` from `someauthor`, I will create the file `src/groups/someauthor.dhall`.
```

PR #402, opened 2019-07-14
Generate packages dhall release

justinwoo:
see packages.dhall in release:

https://github.com/justinwoo/package-sets/releases/tag/test-release-no-verify

example

https://github.com/justinwoo/test-package-sets-releases-packages-dhall

PR #403, opened 2019-07-15
Global warming: remove freezing of package groups

f-f:
This is no longer necessary now that we have #402 and https://github.com/spacchetti/spago/pull/322


justinwoo:
rip remote importers

f-f:
Everything should be fine until we cut a release, so let's hold that until after we get a release for Spago

Issue #32, opened 2016-11-12
Include checksums for each package


Pauan:
This is super-low priority, but I think it would be nice to eventually have checksum information embedded into the `packages.json`

Why? Imagine that the repository `https://github.com/foo/foo.git` is added to the package set with the tag `v1.0.0`

At first, everything is fine. But then later the repository `https://github.com/foo/foo.git` adds in some malicious code and overwrites the `v1.0.0` tag to point to the malicious code. Now everybody who uses that package will end up downloading the malicious code.

But if the checksum for the repository was added to the package set, then `psc-package` could check the checksum after downloading the package `foo`, which would prevent that attack.

Obviously this relies upon `foo` not including any malicious code in the first place, but at least that can be solved by code audits. After a package has been code audited, and the checksum is in place, it is essentially impossible for a malicious change to occur.

(As a side bonus, this also catches any weird network or file system errors that might corrupt the package during download/installation)

On the other hand, if code audits rarely occur, then a checksum could give a false sense of security.

paf31:
👍 Sounds like a good idea.

Pauan:
Also, for the *even more* paranoid, you could create a checksum for the package set itself (including the `packages.json`). This checksum would be placed into the `psc-package.json` file as a new `checksum` field (similar to `set` and `source`)

That means that even if somebody malicious gained access to the package set repo (e.g. `purescript/package-sets`), they *still* wouldn't be able to distribute malicious code, because the package set itself would be checksummed.

Of course `psc-package init` would automatically add in the appropriate `checksum` field, so this would be transparent for users, with no loss of convenience.

Now the entire system (both package set and individual packages) is protected from malicious attacks, so the only parts that need to be trusted are `psc`, `psc-bundle`, and `psc-package` (which can be locally verified, if somebody wished to do so).

I don't think that maliciousness would be a problem for `purescript/package-sets`, but it may be more important if there are alternate package sets created by third parties. I can imagine a third-party package set with a very strict code audit process, which would advertise itself as a "safe alternative to `purescript/package-sets` for enterprises that require maximum safety". Kind of like the OpenBSD of PureScript.

In any case, checksums do not guarantee security, because security *always* requires code audits. And I'm *not* suggesting that `purescript/package-sets` should have strict code audits. Instead, it is more about *enabling* strict code audits, if a third party package set wished to do so.

Also this entire idea is very low priority, as I already mentioned.

sudhirvkumar:
@Pauan 

[...]

PR #426, opened 2019-08-07
Add rev and sha256 fields to packages

jmackie:
This is a first step in adding more integrity checks to the git-based tooling (namely `psc-package` and `spago`).

Specifically I'd like to move everything towards:

1. Using git revisions rather than tags
2. Verifying repo contents via a checksum
3. And maybe also providing a checksum for the package sets themselves (too paranoid?)

Hence, this PR adds `rev` and `sha256` fields to the existing package definitions. 

See #32 for discussion.

### DIFF
--- a/src/groups/ad-si.dhall
+++ b/src/groups/ad-si.dhall
@@ -5,5 +5,9 @@
         "https://github.com/ad-si/purescript-result.git"
     , version =
         "v1.0.3"
+    , rev =
+        "a09ebc3749e3b902f8144fa4a22c95840ce46701"
+    , sha256 =
+        "18s19kp8n659fx36q46qrn2imks43mqs3k7y96cxwdl60smj74pk"
     }
 }
--- a/src/groups/ajnsit.dhall
+++ b/src/groups/ajnsit.dhall
@@ -17,5 +17,9 @@
         "https://github.com/ajnsit/purescript-concur.git"
     , version =
         "v0.3.8"
+    , rev =
+        "f06ff02d3e519db6ceecf4e75a99adca10104695"
+    , sha256 =
+        "0d8idglmz5wsg14rpmmpc54wk6hjfw219mmyjm9fgds5d0p2dhkb"
     }
 }
--- a/src/groups/alexadewit.dhall
+++ b/src/groups/alexadewit.dhall
@@ -11,5 +11,9 @@
         "https://github.com/AlexaDeWit/purescript-text-encoding.git"
     , version =
         "v0.0.9"
+    , rev =
+        "46d809d5806ed9bb725192a7a0cf0d679911aa39"
+    , sha256 =
+        "0zb4cs2wcn2lcfc80gghanclbxy9rgz836dgqs4finj3yg872p47"
     }
 }
--- a/src/groups/anttih.dhall
+++ b/src/groups/anttih.dhall
@@ -5,5 +5,9 @@
         "https://github.com/anttih/purescript-rationals.git"
     , version =
         "v5.0.0"
+    , rev =
+        "8c52d8cc891d1223150a31416220aa9b99404442"
+    , sha256 =
+        "1idvjvvx5kwmi8kj2ps95bcvlsgij1xgin4jfw3rmcqd930wqq6q"
     }
 }
--- a/src/groups/athanclark.dhall
+++ b/src/groups/athanclark.dhall
@@ -5,5 +5,9 @@
         "https://github.com/athanclark/purescript-float32.git"
     , version =
         "v0.1.1"
+    , rev =
+        "1ca78fedb0b2da65b9ec0969d16c0bb7155e4f29"
+    , sha256 =
+        "0g0yi4h9bdc1v8l7my2x19xqas5hbbkkhbj6z4r76559c9rbyvfr"
     }
 }
--- a/src/groups/bodil.dhall
+++ b/src/groups/bodil.dhall
@@ -5,6 +5,10 @@
         "https://github.com/bodil/purescript-signal.git"
     , version =
         "v10.1.0"
+    , rev =
+        "58e44e7f8c20e9d057bcaffbd231893935548bb5"
+    , sha256 =
+        "03jkc6l01vv28drg1rjx9ymc2bndx9933ckf60bi4ar909kx64v5"
     }
 , sized-vectors =
     { dependencies =
@@ -20,6 +24,10 @@
         "https://github.com/bodil/purescript-sized-vectors.git"
     , version =
         "v3.1.0"
+    , rev =
+        "ddb8c8e21a2c37e611a9d031ffbb68b261380851"
+    , sha256 =
+        "1dldwalx2smij4mhhh63qf45ylj1sji029ayppinc3py035shblw"
     }
 , smolder =
     { dependencies =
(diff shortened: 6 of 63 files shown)


f-f:
We are definitely interested in checksums for packages, but I'd say introducing them is quite tricky, as there are several different aspects involved:
1. **picking a hashing algo (this PR)**:
  I'm not sure we want to standardize on Nix's algorithm, as we don't have the explicit goal of being compatible with Nix out of the box (it's easy enough to write wrapper tools, e.g. `spago2nix`, etc), and as far as I know it introduces unnecessary complexity (the whole NAR thing)
  The second problem about this is that we're not going to deal with git repos only, but we need to support tarballs too (and in fact `spago` does fetch tarballs most of the time)
  So in the end I'd say we want some algorithm that can literally be implemented in bash and it's insensitive to the presence of the `.git` folder
2. **getting the data into `package-sets` (also this PR)**
3. **keeping the data up to date (not in this PR, maybe it should be)**:
  i.e. fix tooling for contributors adding new packages or modifying new ones. This includes:
   - CI checks that check that the hash in the `rev` is indeed the hash the tag points to
   - changes to the `make` setup so that it'd go through the group files and add the checksums for the missing packages
  These checks should be implemented in `spago` or in `spago-curator` because some infrastructure for Dhall AST manipulation is there already. I can detail more on this if you wish
4. **checking the data when downloading packages**:
  that is, verifying that the checksum is correct client-side - I can take care of this for `spago`, and I think we don't even want to patch psc-package to use this

A final note about this is that we are already checking the `sha256` of the package-sets itself (in the form of Dhall integrity checks) client-side, so IMHO there is no need to include a checksum of the package set itself

jmackie:
Cheers for the comments m'dude. A few amicable responses...

1. My motivation for using Nix's hashing approach wasn't compatibility (although that's a nice consequence), it was that it seems like a pretty robust solution to the problem. "Hashing a directory" isn't actually that trivial -- see the many stack overflow answers as well as the discussion at the bottom of p.92 https://nixos.org/~eelco/pubs/phd-thesis.pdf. The NAR thing might not be _unnecessary_ complexity. 

Also I'm not sure why we want to support tarballs, other than as an optimization that kind of undermines this integrity checking goal?

4. Why would we not want to patch `psc-package` to do this? It's still the "blessed" package management solution (given its namespace) so should surely follow "best practices" (assuming that's what this becomes)?

And, as I understand it, the way we're currently using Dhall doesn't give us the security assurance I'm after. What I'd like to have is a checksum committed to the repo that users/tools check against, rather than Dhall checking that an expression matches a hash it generated previously. But correct me if I'm wrong here.

f-f:
> Also I'm not sure why we want to support tarballs, other than as an optimization that kind of undermines this integrity checking goal?

So you see, this is why I mentioned this is a tricky problem. An approach that doesn't support tarballs is going to be super slow for clients, so that's kind of a dealbreaker IMO. And I think in the end we will want to just have a package archive where we upload tarballs with checksums, so we might want to hold on this and just do the package registry. That also comes with a set of problems, e.g. [see here](https://discourse.purescript.org/t/blogged-thoughts-on-purescript-package-management/809)

> Why would we not want to patch `psc-package` to do this? It's still the "blessed" package management solution (given its namespace) so should surely follow "best practices" (assuming that's what this becomes)?

As far as I know it is not "blessed" but just happens to be under the `purescript` org, which doesn't mean much (the "blessed" recommendation is still `pulp` + `bower`)

Also as far as I know at this point `psc-package` is considered "stable" and is not receiving upgrades, features, etc. (this is also the reason why `spago` exists)

> And, as I understand it, the way we're currently using Dhall doesn't give us the security assurance I'm after. What I'd like to have is a checksum committed to the repo that users/tools check against, rather than Dhall checking that an expression matches a hash it generated previously. But correct me if I'm wrong here.

Could you detail more the security model you're after and which attack vectors you're considering for the current approach vs what you propose?

jmackie:
> An approach that doesn't support tarballs is going to be super slow for clients, so that's kind of a dealbreaker IMO

I might be missing something here, a quick `time` gives me

```
$ time git clone --depth 1 -b v4.1.1 https://github.com/purescript/purescript-prelude
...
real	0m1.461s
user	0m0.069s
sys	0m0.068s

$ time curl -O https://github.com/purescript/purescript-prelude/archive/v4.1.1.tar.gz
...
real	0m0.709s
user	0m0.036s
sys	0m0.008s
```

So I wouldn't say cloning is _super_ slow (particularly if clones are happening concurrently) and not a deal breaker IMO, particularly if it brings other benefits.

> so we might want to hold on this and just do the package registry

I get the impression from folks on the purescript FP slack (cc @natefaubion) that there's not a desire to do this - the maintenance cost of hosting a package registry is obviously non-trivial and probably not worthwhile. And sticking with a purely git-based approach is appealing for its simplicity (and worth noting that Go seems to be doing fine with this approach). But having said that I'd gladly get behind an effort to create a package registry.

> Also as far as I know at this point psc-package is considered "stable" and is not receiving upgrades, 

Didn't realize that was the case. Should this be mentioned in the README (cc @justinwoo) ?

> Could you detail more the security model you're after and which attack vectors you're considering for the current approach vs what you propose?

Damn you caught me. I don't really know what I'm talking about here. 

Basically I'd like to know that the package-set as a whole and the packages within it are legit when I consume them. Where "legit" means they contain the same code that authors and reviewers have seen and approved. Granted this is only as effective as the review process, which can't feasibly be that thorough right now, but someone could conceivably create a `paranoid-package-sets` using this information.

f-f:
> So I wouldn't say cloning is super slow (particularly if clones are happening concurrently) and not a deal breaker IMO, particularly if it brings other benefits.

In Spago we have to clone the whole repo (and not just shallow clone, so it's much slower for big repos) because you might be referring to a commit hash. `psc-package` doesn't support commit hashes at all.

> Basically I'd like to know that the package-set as a whole and the packages within it are legit when I consume them. Where "legit" means they contain the same code that authors and reviewers have seen and approved. Granted this is only as effective as the review process, which can't feasibly be that thorough right now, but someone could conceivably create a `paranoid-package-sets` using this information.

I'd recommend to read about [Dhall's safety guarantees](https://github.com/dhall-lang/dhall-lang/wiki/Safety-guarantees) for an overview of how the hashing mechanism works and what kind of guarantees you get.
Basically once we have checksums here in `package-sets`'s `master` branch we can assume the package and its contents are trusted and audited and whatever (we don't do this, but we could. Or as you said you could have your own `audited-package-sets` repo).
Then when it comes to consuming the `packages.dhall` from here, you add it to your Spago config and you get a hash in your file. You can then verify that the content is the same you viewed here on the repo by just grabbing the file here and running `dhall hash` on it - in this way you're sure the contents of the file you saw and the one Spago downloads are the same. Having the hash in-band (i.e. in the file) would only be additional work but would not give you additional guarantees (the git commit hash is kind of already guaranteeing this if you squint) since you'd have to check _anyways_ that the hash of the file itself is the same of the one you saw.

justinwoo:
I will close this PR as there is no current plan to do anything with it, and it interferes with my maintenance work on this repo. Please comment on the linked issue https://github.com/purescript/package-sets/issues/32 for further discussion. Thanks.

PR #435, opened 2019-08-14
Add nkly/purescript-express to package set

drewolson:


justinwoo:
thanks. this package was removed previously since it didn't update its usage of Foreign Generic in over a month, but this is fine now it seems.

PR #488, opened 2019-09-18
Update Radox to v1.0.0 and add Aff dependency

danieljharvey:
Hello!

Hope all is well! Just a version bump here.

Thanks
Dan


f-f:
Thank you! 🙂

PR #497, opened 2019-09-26
Update node-buffer to v6.0.0

Dretch:
This won't compile until https://github.com/purescript/package-sets/pull/496 merged.

CC @hdgarrood


justinwoo:
Try rebasing now?

PR #500, opened 2019-09-28
Add .git ending to unsafe-reference repo link

bjornkihlberg:
Make unsafe-reference be used with spago

[Issue-499](https://github.com/purescript/package-sets/issues/499)

Issue #499, opened 2019-09-28
missing .git for unsafe-reference seems to make it unusable from spago


bjornkihlberg:
The [unsafe-reference](https://github.com/purescript-contrib/purescript-unsafe-reference) package doesn't seem to work with [spago](https://github.com/spacchetti/spago) because the repo link is missing a **.git** at the end.
```json
  "unsafe-reference": {
    "dependencies": [],
    "repo": "https://github.com/purescript-contrib/purescript-unsafe-reference",
    "version": "v3.0.1"
  },
```

justinwoo:
Fixed in #502

PR #502, opened 2019-10-03
Fix url of unsafe-reference

jhrcek:
This is proper fix for https://github.com/purescript/package-sets/issues/499 which supersedes https://github.com/purescript/package-sets/pull/500 in which author fixed the url in package.json, but not in the dhall config.


justinwoo:
cool, thanks

PR #514, opened 2019-10-10
Fix URL of purescript-spork

jhrcek:
Spago gives warning and refuses to cache this repo locally, because the URL doesn't have the required form. While I was at it I checked all the urls in the package set and all the others have the correct form.


f-f:
Thanks!

PR #541, opened 2019-11-25
update to 0.13.5

justinwoo:


justinwoo:
eventually the github outage will be over and we will be able to see what travis thinks about this PR

justinwoo:
nice, it actually started running before the outage

PR #542, opened 2019-11-27
Fix concur-react and bump versions

ajnsit:
Add concur-core which was split out into its own repository. Bump
concur-react version. Point to updated repo locations.


justinwoo:
thanks

PR #552, opened 2020-01-15
Add purescript-interpolate

JordanMartinez:
Before merging this, please be aware of these two things.

First, ["Registering bower package names is not supported anymore"](https://github.com/JordanMartinez/purescript-interpolate/issues/1#issue-548595305) (related issue: purescript/pursuit#402)

Second, this does not work...
```bash
$> bower i purescript-interpolate
bower ENOTFOUND     Package purescript-interpolate not found
```
... but this does

```bash
$> bower i jordanmartinez/purescript-interpolate
```

If there are other libraries that build upon this one, I'm not sure whether the above facts will be an issue or not as `bower i -p` seems to work.

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -1535,6 +1535,13 @@
     "repo": "https://github.com/purescript/purescript-integers.git",
     "version": "v4.0.0"
   },
+  "interpolate": {
+    "dependencies": [
+      "prelude"
+    ],
+    "repo": "https://github.com/jordanmartinez/purescript-interpolate.git",
+    "version": "v2.0.1"
+  },
   "invariant": {
     "dependencies": [
       "prelude"
--- a/src/groups/jordanmartinez.dhall
+++ b/src/groups/jordanmartinez.dhall
@@ -1,0 +1,6 @@
+{ interpolate =
+    { dependencies = [ "prelude" ]
+    , repo = "https://github.com/jordanmartinez/purescript-interpolate.git"
+    , version = "v2.0.1"
+    }
+}
--- a/src/packages.dhall
+++ b/src/packages.dhall
@@ -78,5 +78,6 @@ let packages =
       ⫽ ./groups/ebmtranceboy.dhall
       ⫽ ./groups/purescript-concur.dhall
       ⫽ ./groups/csicar.dhall
+      ⫽ ./groups/jordanmartinez.dhall
 
 in  packages


f-f:
@JordanMartinez yeah I followed the above threads as they were happening.

Two notes:
1. Given that the Bower registry has changes their policy on registering names, I think we should lift the restriction on "the package must be published on the Bower registry", and change it to "the package must have a valid `bower.json` file" or "the package can be installed with Bower" (cc @justinwoo @hdgarrood)
2. Given the above point, we should totally merge this

f-f:
@hdgarrood side note: we did discuss the possibility of a PureScript registry before, and if the "Bower policy change" that just happened changed your stance on this I'd love to have a chat about it (of any form - can be a GitHub thread, a Slack conversation, a videocall, anything really)

hdgarrood:
Yeah, we definitely need to come up with a new policy here. I’d like to discuss on discourse - since this issue kind of cuts across the whole ecosystem I feel like discussing on one individual github repo kind of feels inappropriate.

hdgarrood:
More recently I’ve been feeling that rather than continuing to try to have a centralised name registry we should instead embrace the distributed nature of package sets and just start naming packages in such a way that tells you where you can get them from. So for example we might want to start saying github:purescript/purescript-prelude or something. I think both elm and entropic use schemes like this already. I haven’t thought through the details yet though.

f-f:
@JordanMartinez we just updated publishing requirements in #577, and this package is now ["published"](https://github.com/purescript/registry/blob/74d225b11ec35547ef274c3b3cbcfcb6382dcc1b/new-packages.json#L2) by the new definition, could you rebase this PR?

JordanMartinez:
Thanks for the notification. I rebased and ran `make && cd src/ && spago verify interpolate` before force-pushing. Looks like CI failed because the package.json file changed (and it seems like it just sorted dependencies by alphabetical order).

f-f:
@JordanMartinez it looks like your PR has shuffled some things in the json file, so CI rightfully complains about that.
Most likely you have outdated versions of the tools CI is using (e.g. `dhall-to-json`, `psc-package`), could you try running `nix-shell --command "make"`?
This will format the file using the same versions as in CI (which runs `make ci` in the same `nix-shell`)

JordanMartinez:
Thanks. I downloaded a newer version of dhall and dhall-to-json and that fixed the CI issue.

f-f:
Looks great, thank you! :slightly_smiling_face:

PR #575, opened 2020-03-04
Updates 2020-03-04

spacchettibotti:
Updated packages:
- [`argonaut-core` upgraded to `v5.0.2`](https://github.com/purescript-contrib/purescript-argonaut-core/releases/tag/v5.0.2)
- [`polyform` upgraded to `v0.8.2`](https://github.com/paluh/purescript-polyform/releases/tag/v0.8.2)
- [`st` upgraded to `v4.1.1`](https://github.com/purescript/purescript-st/releases/tag/v4.1.1)

You can give commands to the bot by adding a comment where you tag it, e.g.:
- `@spacchettibotti ban react-basic`
- `@spacchettibotti unban simple-json`

PR #576, opened 2020-03-06
Add int-53, redis-client and redis-hotqueue

paluh:
Unfortunately I wasn't able to run `psc-package verify` (I'm getting enigmatic `does not exist`). On the other hand I'm using this version package-sets and the above packages in my current project it and it compiles without any problems :-)

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -1516,6 +1516,15 @@
     "repo": "https://github.com/garyb/purescript-indexed-monad.git",
     "version": "v1.2.0"
   },
+  "int-53": {
+    "dependencies": [
+      "integers",
+      "math",
+      "strings"
+    ],
+    "repo": "https://github.com/rgrempel/purescript-int-53.git",
+    "version": "v4.0.0"
+  },
   "integers": {
     "dependencies": [
       "globals",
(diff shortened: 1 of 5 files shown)


f-f:
@paluh thank you for this! It seems like your latest commit is missing some packages that are required to build the new ones you added.

And thanks for noting that our instructions were outdated (`psc-package` had a breaking change in https://github.com/purescript/psc-package/pull/158), I opened #577 to fix the issue and switch the instructions to use Spago instead

paluh:
@f-f Thanks for a lot for your feedback and sorry that I've missed out that my second commit had broke the CI build.

 > I opened #577 to fix the issue and switch the instructions to use Spago instead

It is really good to know. Thanks!

I'm only not sure what the current status of this PR is. I'm bit lost regarding your last commit which is visible above. Could you please tell me if it is included already in this PR and fixes the issue with missing dependencies or should I push the fix myself?

f-f:
@paluh if you're done adding packages then this PR is good to go. Can I merge?

The reason why your commit broke CI is that you upgraded `polyform` to `0.8.2`, which included a dependency on `validation`, which the previous version `0.8.0` did not include. I fixed this independently in #575, which is why merging `master` in here fixed the build

paluh:
> (...)￼ if you don't have any more packages to add then this PR is good to go. Can I merge?

Yeah, sure. Thanks!

P.S.
I'm going to push more of our "core" libs like postgresql-client, selda etc. but I will open a separate PR fot this.

f-f:
Thank you! :slightly_smiling_face:

PR #604, opened 2020-04-19
arraybuffer-class, zeta, and an update to uint

athanclark:
The dependencies iterated in these packages might be excessive - is there a way to declare "dev-dependencies" in spago / psc-package? If so, I'll likely publish a revision to these later.


athanclark:
Sorry, I just realized how much of a hot mess this is. I'll try again later.

PR #606, opened 2020-04-20
Add addaybuffer-class, monad-control, z85

athanclark:
Hi, I'm not sure if it is normal for `make` to touch many different .dhall files in groups/, but it did, that's why I separated the two commits.

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -3961,15 +3961,5 @@
     ],
     "repo": "https://github.com/athanclark/purescript-z85.git",
     "version": "v0.0.2"
-  },
-  "zeta": {
-    "dependencies": [
-      "aff",
-      "arrays",
-      "foreign-object",
-      "refs"
-    ],
-    "repo": "https://github.com/athanclark/purescript-zeta.git",
-    "version": "v5.0.4"
   }
 }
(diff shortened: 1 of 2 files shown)


f-f:
It looks like you're running an old version of `dhall` or `dhall-to-json`, could you try running `nix-shell --command 'make'` instead? This will pick up the right versions for you.

athanclark:
Will do, surprising since I ran `stack install dhall` etc.

athanclark:
@f-f Hey sorry about that, nix-shell worked without a hitch. I rebased the last commits so there's no format overwriting. Thanks!

athanclark:
Shoot, this build is failing - I tried to design `zeta` as a replacement for `signal`, meaning the module names were overwritten. I didn't realize that a package set is meant to guarantee every package can work with one another. Should I redesign `zeta` against a different module name?

f-f:
@athanclark yes, module names in the package set must be unique because as you noted all the packages in the set have to work together. So yeah, changing module name is the only possiblity here if you'd like the package in the set

athanclark:
@f-f Sounds good. I commented out the zeta package for now in my athanclark.dhall file - I'll submit a separate pull request when it's ready, if that's okay. Everything else seems to be working a-ok though!

PR #617, opened 2020-05-01
Drop strongcheck

thomashoneyman:
Strongcheck has [been deprecated](https://github.com/purescript-deprecated/purescript-strongcheck), so it probably ought to be dropped from package sets as well. All other -contrib deprecations were already not in package sets.

If this causes conflicts we can leave the package in the set until it goes out of date.

### DIFF
--- a/src/groups/purescript-contrib.dhall
+++ b/src/groups/purescript-contrib.dhall
@@ -216,39 +216,6 @@
   , repo = "https://github.com/purescript-contrib/purescript-strings-extra.git"
   , version = "v2.1.0"
   }
-, strongcheck =
-  { dependencies =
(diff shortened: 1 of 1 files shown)


f-f:
Thanks @thomashoneyman! Usually the policy is that we don't drop packages from the set until they cause trouble, but I'm also fine with this since there's an explicit deprecation.

CI is failing because you should run `nix-shell --run 'make'` and commit the `packages.json` too

f-f:
Nvm, I'm adding this change in #616. I'll close this :slightly_smiling_face:

thomashoneyman:
Sorry @f-f, I keep forgetting about running `make`.

PR #634, opened 2020-05-21
Add tree-rose

JordanMartinez:
`purescript-tree-rose` is a fork of the currently unmaintained `purescript-tree` library. `purescript-tree` isn't yet in the package set, but I created a new name that's more specific for what kind of tree this is. I also created a new repo rather than using my fork because I recall reading that some issues can arise with forked repos when the original repo is deleted or something. I'm not sure whether that is still true, but I thought "better safe than sorry."

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -3784,6 +3784,15 @@
     "repo": "https://github.com/purescript/purescript-transformers.git",
     "version": "v4.2.0"
   },
+  "tree-rose": {
+    "dependencies": [
+      "free",
+      "lists",
+      "prelude"
+    ],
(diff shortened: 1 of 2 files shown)


f-f:
@JordanMartinez thank you! CI is failing because it needs the updated `packages.json` too, could you run `nix-shell --command make`?

f-f:
@JordanMartinez also I think this will need to be added to the `registry` as well (since it's a "new repo")?

JordanMartinez:
Ah... sorry about that. I forgot. I've ran the command and submitted a PR to the registry. purescript/registry#25

PR #643, opened 2020-06-20
Add web-url

mjepronk:
Please add purescript-web-url to the package set. Thank you!


f-f:
Note that while you added a new group file, the package was not added to the set, as you'd need to add a new import [to this file](https://github.com/purescript/package-sets/blob/master/src/packages.dhall)

mjepronk:
@f-f I'm sorry for my incomplete PR Fabrizio, I've added the group file to `packages.dhall` and also commited `packages.json`.

PR #644, opened 2020-06-20
Move CI to GitHub Actions

f-f:
..as Travis doesn't seem to like Nix anymore.


abathur:
@f-f It's probably not relevant for you anymore, but just wanted to let you know that Nix support is working on Travis again as of about an hour ago.

f-f:
@abathur thanks for the update! I kind of did this in anger as I needed to just get CI to work and I had no idea if I'd like GitHub actions in the long run, but they seem to be pretty cool actually, so I think we'll stick to them for a while :slightly_smiling_face:

Issue #642, opened 2020-06-19
Managing argonaut-codecs update


thomashoneyman:
The `argonaut-codecs` package has been updated to `v7.0.0`, which introduces typed errors. A number of libraries in the ecosystem depend either on `argonaut` or `argonaut-codecs`, which I have listed below.

All of the libraries below have links to their open pull requests which update them, or if they've already been updated, the links point to their release which includes the change.

I'd like to avoid updating argonaut-codecs and these libraries until most or all of them have had the chance to update so that we can have a graceful move from one package set to the next. But if that's not a reasonable ask, then I'm open to taking another approach.

I don't know if a change has propagated through a lot of libraries in the package set before, so I'm not sure if I'm handling this in the best way. Maybe this can be a nice warmup for all the changes that will happen with PureScript 0.14!

I'm open to advice on how to make sure this is an effective transition.

- [x] [purescript-contrib/argonaut](https://github.com/purescript-contrib/purescript-argonaut/releases/tag/v7.0.0)
- [x] [purescript-contrib/argonaut-generic](https://github.com/purescript-contrib/purescript-argonaut-generic/releases/tag/v6.0.0)
- [x] [purescript-contrib/argonaut-traversals](https://github.com/purescript-contrib/purescript-argonaut-traversals/releases/tag/v8.0.0)
- [x] [matthew-hilty/struct](https://github.com/matthew-hilty/purescript-struct/pull/2) (PR still open, but library is compatible with both Argonaut versions)
- [x] [matthew-hilty/tolerant-argonaut](https://github.com/matthew-hilty/purescript-tolerant-argonaut/releases/tag/v2.0.0)
- [x] [flip111/ordered-set](https://github.com/flip111/purescript-ordered-set/pull/1)
- [ ] [joneshf/option](https://github.com/joneshf/purescript-option/pull/22)
- [x] [thomashoneyman/slug](https://github.com/thomashoneyman/purescript-slug/releases/tag/v2.0.0)
- [x] [purescript-hyper/hyper](https://github.com/purescript-hyper/hyper/releases/tag/v0.11.1)
- [x] [purescript-hyper/purescript-trout](https://github.com/purescript-hyper/purescript-trout/releases/tag/v0.12.3)
- [x] [purescript-hyper/purescript-trout-client](https://github.com/purescript-hyper/purescript-trout-client/releases/tag/v0.12.1)
[...]

PR #651, opened 2020-07-01
Update for argonaut-codecs

thomashoneyman:
This PR updates the package set to accommodate the change to argonaut-codecs to introduce typed errors.

Currently, these libraries are being removed from the package set. As noted in #642 each of these (except trout-client, which is waiting on an update to hyper) have had pull requests open for 11 days at the time this PR was opened.

- [ ] option (@joneshf)
- [ ] polyform-validators (@paluh)
- [x] refined (@danieljharvey)
- [x] tolerant-argonaut (@matthew-hilty)
- [x] trout-client

This isn't urgent to merge because most folks are on the older package set. However, Bower users are going to be installing incompatible versions of the other ~15 libraries that have been updated, so it's still good to merge this at some point in the next week or two.


danieljharvey:
Have just released v1.0.0 of `refined` that uses the new Argonaut.

thomashoneyman:
Thanks! There are now only 4 libraries in the package set affected by the change.

thomashoneyman:
We're down to 3 libraries affected. I suggest we merge this in a week (next Friday), but I understand if we need to wait a little longer to try and get _all_ the libraries in.

thomashoneyman:
@f-f The only remaining libraries affected are polyform-validators (@paluh) and option (@joneshf). I think this should be merged this Friday, but I am open to pushing it back another week.

f-f:
@thomashoneyman nothing has moved in a week, so I'd say we could go ahead with this?

thomashoneyman:
Yes, let’s do it!

PR #652, opened 2020-07-02
Update react-basic packages

spicydonuts:
Hopefully this is correct, I haven't done this before. Running the commands in the docs reformatted everything, so I split the PR into two commits: 1, the react-basic updates and 2, the reformatting.


f-f:
Oh, it looks like there's some outdated docs somewhere then, could you point me to them?

Nowadays the recommended way to do this is to run `nix-shell --run 'make ci'`, which should not reformat.
Though I'm not quite sure why CI doesn't fail for the reformatting.. :thinking:

spicydonuts:
This: https://github.com/purescript/package-sets/blob/master/CONTRIBUTING.md#how-to-add-a-package-to-the-set

I've never used nix and the nix installer failed, so I just installed the dependencies. That guide doesn't mention `make ci`, just `make` and `cd src && spago verify [new-package]`.

spicydonuts:
Just ran `make ci` successfully

f-f:
Oh, I wrote `make ci` but I actually meant to write `make`, sorry for the trouble :see_no_evil: 

Thanks for the link to the outdated docs, I'll fix them up to point to Nix or I'll specify the versions there.
In fact, the reformatting is happening because your versions of `dhall` and `dhall-to-json` don't match the ones that CI is using - in the `nix-shell` we have these versions:
- `dhall`: `1.32.0`
- `dhall-to-json`: `1.6.4`

I think we have mainly two ways forward here:
- you can install the specific versions above
- or we can try to debug your `nix` install, so that you can just run `nix-shell`

spicydonuts:
I reformatted my branch in nixos docker container. Does this look better?

f-f:
Yes this looks great thanks! :clap: 

One last thing we need is that `react-basic-classic` doesn't seem to be on the registry, so I opened a PR for it: https://github.com/purescript/registry/pull/40
I'll wait for you to take a look before merging that.
The background on why that's needed is that we need to note the name-to-package association somewhere now that we cannot publish on the Bower registry. See [here](https://github.com/purescript/package-sets/blob/master/CONTRIBUTING.md#general))

Could you also share some instructions on how you did this with the nixos container so that we can include that in the docs?

spicydonuts:
All I did was `docker run -it nixos/nix /bin/sh`, then `nix-shell -p git`, copy/paste my ssh keys, and clone/push. If it's added to the docs here it should probably be improved to run as a one-liner that mounts the current directory inside the container just to run `make ci`, which is a bit more involved.

spicydonuts:
This seems to work: `docker run -t --mount type=bind,source="$(pwd)",target=/package-sets nixos/nix /bin/sh -c 'cd /package-sets; nix-shell --run "make && make ci"'`

PR #653, opened 2020-07-04
Add halogen-svg-elems

JordanMartinez:
This will fail to build until purescript/registry#38 is merged.


f-f:
@JordanMartinez CI will not fail to build if the package is not in registry, the check is manual so far :sweat_smile:

PR #663, opened 2020-07-24
react-markdown, zeta-array, and zeta-mapping

athanclark:


f-f:
It looks like the packages you're adding here have not been added to [the registry](https://github.com/purescript/registry/blob/master/new-packages.json), could you add them first?

f-f:
This is now very out of date, so I'll close it

PR #679, opened 2020-08-30
Add packages from zudov

thomashoneyman:
This PR adds the `websocket-simple` and `var` packages by @zudov. They are already published on Bower and this adds them to the package sets as well.

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -4289,6 +4289,15 @@
     "repo": "https://github.com/purescript/purescript-validation.git",
     "version": "v4.2.0"
   },
+  "var": {
+    "dependencies": [
+      "contravariant",
+      "effect",
+      "invariant"
+    ],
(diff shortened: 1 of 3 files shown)


thomashoneyman:
Uh oh:

```
Error found:
at .spago/websocket-moderate/v7.0.2/src/WebSocket.purs:1:1 - 33:28 (line 1, column 1 - line 33, column 28)

  Module WebSocket has been defined multiple times


See https://github.com/purescript/documentation/blob/master/errors/DuplicateModule.md for more information,
or to contribute content related to this error.
```

I'm not sure the best thing to do about this. We don't have a proper solution for namespacing these.

f-f:
These packages can be safely regarded as unmaintained since long, so we should probably just fork them if we think there's any value in the code

thomashoneyman:
Understood. I'm going to close this PR with that in mind.

PR #680, opened 2020-08-31
Add gl-matrix

dirkz:
Add a purescript wrapper to glmatrix.net

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -1356,6 +1356,18 @@
     "repo": "https://github.com/purescript/purescript-generics-rep.git",
     "version": "v6.1.1"
   },
+  "gl-matrix": {
+    "dependencies": [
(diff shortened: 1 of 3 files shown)


f-f:
It looks like CI is failing because the packages you're adding here have not been added to [the registry](https://github.com/purescript/registry/blob/master/new-packages.json), could you add them first?

dirkz:
Thanks for the heads up. Will add it to the registry, then remove test dependencies from the build, then try again.

thomashoneyman:
@dirkz you should now be able to make that update

dirkz:
Updated the lib to v2.0.1, now without quickcheck. All good from my side.

f-f (approved):

PR #694, opened 2020-09-11
Add murmur3

ursi:


ursi:
oops, I thought if it installed it was okay to use. Guess I gotta try to compile it first.

PR #702, opened 2020-09-18
Add latest-compatible-sets.json file

### LINKED ISSUE #701: Store mapping of compiler versions to latest compatible package sets
For example, we might have a file in the repository root `latest-compatible-sets.json` whose contents might look like this:
```
{
  "0.13.8": "psc-0.13.8-20200911-2",
  "0.13.7": "psc-0.13.6-20200507",
  "0.13.6": "psc-0.13.6-20200507",
  [...]
  "0.12.5": "psc-0.12.5-20190525"
  [...]
}
```

This would be useful for https://github.com/purescript/spago/issues/657, and potentially other tools as well?

This seems like the kind of thing we might want to automate, although I'm not quite sure how to go about that just yet.

hdgarrood:
This file is automatically updated when new releases are created using
the new update-latest-compatible-sets.sh script. Fixes #701 

You can see this in action by looking at the `master` branch of my fork. For example: https://github.com/hdgarrood/package-sets/commit/334bd91fb30d3018b9cb7e1e56be9eb0af57fbc5


latest-compatible-sets.json
```diff
@@ -0,0 +1,19 @@
+{
+  "0.11.4": "psc-0.11.4",
+  "0.11.5": "psc-0.11.5",
+  "0.11.6": "psc-0.11.6-prime",
+  "0.11.7": "psc-0.11.7-20180524",
```
f-f, line 5:
```suggestion
```

Could we drop support for the package sets for `0.11.x` here?
This file is supposed to be used for starting new projects, so I think we should not encourage `0.11`

f-f (approved):
This looks really great, I left a small comment but otherwise it's good by me :heart:

latest-compatible-sets.json
hdgarrood: Thanks! Dropping support for 0.11.x sounds good to me.

f-f:
Thank you :slightly_smiling_face:

thomashoneyman:
I'm not sure this is working properly in CI: I published a new tag and the build fails:
https://github.com/purescript/package-sets/runs/1151048508

Specifically in the `Commit changes to latest-compatible-sets.json` step which uses the `stefanzweifel/git-auto-commit-action@v4` action.

hdgarrood:
Ah, I didn't realise the master branch in this repository is protected. Maybe rather than using the `git-auto-commit-action` action, we should have a slightly more sophisticated process where the updated JSON file is added to a pull request, which then gets merged automatically once it succeeds?

hdgarrood:
Nobody is relying on this file yet AFAIK, so feel free to comment out the action or revert this change to get the new tag published.

f-f:
Maybe at this point we could just make a release script that does this?

We already have to do a set of annoying things on release:
- figure out how to call the tag
- cut the tag locally and push it (we cannot create a release here on GH because the Release CI assumes there's a tag but no release yet)

hdgarrood:
As in a script which runs locally on the maintainer's machine? I think if the master branch is protected then you still have the same problem, unless you want to push the tagged commit to a different branch to start with (which seems a little dubious to me).

f-f:
The update of the file should happen _after_ the tag is cut, and it's independent from the tag commit, so we can just create a PR after we push the tag.

Going more in detail, I propose that the release script would:
- figure out how to name the tag
- create it
- push it here
- then create a branch
- update the file `latest-compatible-sets.json` file in the branch
- push the branch
- open a PR (with the GitHub API or with the new GitHub CLI) to be merged manually or by `mergify`

PR #706, opened 2020-09-24
Update jamesdbrock.dhall

jamesdbrock:
Upgrade arraybuffer-builder
Add longs (belongs to zapph)
Add protobuf


jamesdbrock:
I can't tell from the tests log why tests are failing?

thomashoneyman:
It looks like it's failing because the `.dhall` file has been updated, but the generated JSON has not. Do you mind running `nix-shell --command make` and committing the result? I believe that should work.

thomashoneyman (approved):

PR #716, opened 2020-10-01
Update CI: test if packages updated to 0.14 rc build without err/warn

JordanMartinez:
This updates the CI for the `prepare-0.14` branch to test whether the packages listed in `src/spago.dhall` compile using the current release candidate of PureScript v0.14.0

I'm not familiar with GitHub Actions yet, so I'm not sure whether this will work the first time. I also am not using Nix because I don't know how to configure it to use the release candidate.


JordanMartinez:
Alright. The CI for this branch now compiles packages into the `src/spago.dhall` file properly now. `prelude` doesn't compile because purescript/purescript-prelude#227 hasn't yet been merged into `master`.

One thing that would make this better is if CI failed when any warnings are emitted. Unfortunately, I'm not sure how that would be done.

JordanMartinez:
To make this build fail when warnings are emitted, I'll need to use [Harry's compatibility update for `purescript-psa`](https://github.com/natefaubion/purescript-psa/pull/44).

f-f (approved):
Looks great! :clap: 

I saw your thread on the Slack about having `--strict` fail the build if there are warnings in the dependencies. Something that could work for our usecase would be to have two Spago configs:
- one with the packages as dependencies and that we'd only use to call `spago install`, so that we'd get the dependencies downloaded in the `.spago` folder
- once we do that we could move all the package sources to `src` (so that `psa` would find out)
- and then compile with `spago build -u "--strict"` with another spago config with no dependencies

Would this work?

JordanMartinez:
@f-f Thanks so much for the idea! Yes, it does work. I got so annoyed at this yesterday that I stopped working on it yesterday.

The latest commit verifies that CI will fail when we compile packages that still have compiler warnings. I'll update `prelude` back to `master` after I make this comment, and then I'll merge it in.

PR #717, opened 2020-10-03
Include multiple `purescript` org repos as updated libraries

JordanMartinez:


src/groups/purescript.dhall
```diff
@@ -97,7 +97,7 @@
 , effect =
   { dependencies = [ "prelude" ]
   , repo = "https://github.com/purescript/purescript-effect.git"
-  , version = "v2.0.1"
+  , version = "master"
```
andys8, line 100: Is this intended?

JordanMartinez: It was, but I'm wondering whether the original approach I was using to update the ecosystem should still be done.

andys8: Comparing it with prelude, I do now understand the intention.

JordanMartinez: I think we'll use `bower` to update each core library and then use `spago` to verify that previous PRs made to core libraries still work on later PureScript rc releases. For example, `prelude` was updated when `v0.14.0-rc2` was the release candidate. Harry made the `-rc3` version now, so does `prelude` still compile on it? We could use `spago` and this branch to track that so we don't need to waste time updating each individual core library whenever a new release candidate is made.

PR #719, opened 2020-10-05
add dodo-printer

3ddyy:


src/groups/natefaubion.dhall
```diff
@@ -116,4 +116,23 @@
...
+  { dependencies =
+    [ "aff"
+    , "ansi"
+    , "avar"
+    , "console"
+    , "effect"
+    , "foldable-traversable"
+    , "lists"
+    , "maybe"
+    , "minibench"
```
kritzcreek, line 129: Those all seem like test dependencies:

- `console`
- `minibench`
- `node-child-process`
- `node-fs-aff`
- `node-process`

which you shouldn't add here (they can live in the `spago.dhall` file inside the project instead).

natefaubion:
Thanks for doing this! The only src dependencies are:
```
  [ "ansi"
  , "foldable-traversable"
  , "lists"
  , "maybe"
  , "strings"
  ]
```

3ddyy:
Thanks for the dep list! and sorry for the wait

thomashoneyman (approved):

PR #736, opened 2020-10-20
Add release script

f-f:
This adds a `release.sh` script so that we can just run that to get new releases instead of doing things by hand.
It will also update the `latest-release.tags.json`, so I removed the relevant section from CI. However, to allow pushing straight to `master` I unprotected it _only for administrators_. This adds some more responsibility on us (because we can mess up), but all in all I think it's an OK tradeoff


hdgarrood (approved):

.github/workflows/release.yml
```diff
@@ -33,16 +33,4 @@ jobs:
...
-          ref: 'master'
-      - name: Update latest-compatible-sets.json
-        run: |
-          nix-shell --run './update-latest-compatible-sets.sh ${{ github.ref }}'
-      - name: Commit changes to latest-compatible-sets.json
-        uses: stefanzweifel/git-auto-commit-action@v4
-        with:
-          commit_message: Update latest-compatible-sets.json for ${{ github.ref }}
-          file_pattern: latest-compatible-sets.json
+          asset_content_type: application/x-dhall
```
thomashoneyman, line 36:
```suggestion
          asset_content_type: application/x-dhall

```

release.sh
```diff
@@ -0,0 +1,24 @@
...
+
+git add latest-compatible-sets.json
+
+git commit -m "Update to the latest tag"
+
+git push
+
+git tag "${NEW_TAG}"
+
+git push origin "${NEW_TAG}"
```
thomashoneyman, line 24:
```suggestion
git push origin "${NEW_TAG}"

```

thomashoneyman:
The idea is that one of us will run this on a semi-regular basis from our machines?

thomashoneyman (approved):

f-f:
@thomashoneyman yeah, the idea is that we'd run that on our machines whenever we want to cut a release. I agree that it'd be nice for it to list the new packages in the release! I also left a TODO for proper handling of the tag dates, but I'm pretty happy with this so far, as it automates things quite nicely already

PR #745, opened 2020-11-09
Update debuggest

ursi:


ursi:
oops, forgot about the json again

PR #748, opened 2020-11-20
Add midi to the package sets

newlandsvalley:


newlandsvalley:
The first commit of the three is spurious and has been backed out by the last one.  I don't understand why the checks fail.

thomashoneyman:
@newlandsvalley The checks were failing because GitHub [deprecated `add_path` in Actions](https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/) due to a security issue, and we used this in CI. That's now fixed!

thomashoneyman:
I opened #750 to add this package as there are some funky changes to the `package.json` here and I was having trouble fixing that locally. Regardless, the package has been added! Sorry for the delay -- I was out on vacation.

newlandsvalley:
Thank you very much for sorting this out @thomashoneyman.  Yes - very sorry about this.  What had happened is that I'd forked package-sets ages ago to experiment (badly) with adding my module.  What I hadn't realised when I came to it again was that I'd already committed these incorrect changes.  So I struggled a bit and had to try to correct it in the final commit, but the commit history was bizarre!

PR #760, opened 2020-12-11
Update these dependencies

kl0tl:
These (eheh) were added in https://github.com/purescript-contrib/purescript-these/pull/29.


kl0tl:
This should build once https://github.com/purescript-contrib/purescript-these/pull/32 is merged.

PR #767, opened 2020-12-17
Automatically generate packages.json file on pull requests

thomashoneyman:
This PR modifies the GitHub Actions pipeline to update and commit (if necessary) the `packages.json` file.

With this change, contributors only need to update the relevant `.dhall` files (which can often be done strictly through the GitHub UI, minimizing the effort required). If the update changes the packages.json file then it will automatically be committed to their branch. If it does not, then nothing will be committed. Either way the tests will still run -- they don't rely on the packages.json file in any case.

You can see an example of this action working in this commit, performed by the GitHub Actions user: https://github.com/purescript/package-sets/commit/77303d42c6ace57e2d5f04fc9d9856856232ef87

I _believe_ this should work on any pull request to the repository, but I can't test that until this is merged and a PR from an account without access to this repository makes a pull request. So I propose:

1. If this generally looks good then we merge it
2. Another member of the Contributors org who doesn't have access to this repository updates a package
3. We verify that this CI does properly commit to their branch
4. If it does not, then I revert this commit to go back to the previous CI process.

@f-f I'm not sure what README updates to make to help casual contributors know that they just have to update the relevant `.dhall` file for their changes. Do you have any suggestions?

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -2,7 +2,9 @@ name: build
 
 on:
   pull_request:
+    branches: master
   push:
+    branches: master
 
 jobs:
   build:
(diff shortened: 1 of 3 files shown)


thomashoneyman:
Note that the 'Update packages.json.' commit is only re-running tests because I opened a pull request. Actions are not re-triggered when the GitHub Actions user commits to the branch.

PR #770, opened 2020-12-19
Update packages

f-f:


f-f:
@thomashoneyman the automation seems to work well, except the status doesn't seem to be reported :thinking:

thomashoneyman:
Hm. I guess that’s because the required ‘tests’ check doesn’t run on the last commit (the update package) because it’s committed by the action. I’ll have to see how we can get around this.

f-f:
I think running it on every commit would be harmless: once the Action updates the `packages.json` then it would stop triggering the action to add more commits? (Since the diff will be empty)

thomashoneyman:
@f-f it now re-triggers and runs successfully.

thomashoneyman:
I'm making one more update -- now, the full tests will only be run if the packages.json was unchanged (to avoid burning minutes on the lengthy tests when we know that we're committing and re-running again).

thomashoneyman:
This looks good now. We just need to be careful updating these workflows in the future to make sure we never get into a loop.

f-f:
Looks great, thank you @thomashoneyman! :heart:

PR #776, opened 2020-12-20
CI: Create a separate workflow for updating the packages.json file

thomashoneyman:
As noted by @andys8, it's quite convoluted to try and commit from a fork, which is the most common reason our workflows will be triggered in this repository (external contributors adding a package).

Here's the problem: we require that pull requests to master pass the required `tests` check in order to be merge-able. However, this check won't run if the last commit on a pull request was made using the `GITHUB_TOKEN` that workflows receive; if we use that token to commit an updated packages.json file then our tests will not re-run.

We _can_ trigger a workflow again if we make the commit using a personal token, stored in a repository secret. Unfortunately, it turns out that GitHub will not make secrets available to forks (for security reasons), so any pull request from a fork will fail to run checks at all (not only on the last commit).

This PR gets us out of the mess by using one workflow for verifying a user's changes to the `.dhall` files and another workflow for committing a generated package.json to master. We still generate the file as part of testing the user's branch, but we don't commit it to master until their PR has been merged.


thomashoneyman:
@f-f I've merged this for the sake of unblocking existing PRs, but we can iterate on it going forward if needed.

f-f:
@thomashoneyman thank you!

thomashoneyman:
@f-f Yeesh -- unfortunately, I can't push to master:

```
remote: error: GH006: Protected branch update failed for refs/heads/master.
```

f-f:
We could unprotect it..

thomashoneyman:
I think if we were going to unprotect the branch then we'd be better off removing `tests` as a required check (this would happen anyway if the branch was totally unprotected). We'd still as maintainers consider it a 'required' check. Then, we could go back to having the packages.json committed to the pull request itself.

With this solution:

* the master branch is still protected from inadvertent commits
* the packages.json file is still generated, just as part of the PR that introduces a change to the package set
* the tests still run, just not on the packages.json commit (which we don't run tests on anyway)

PR #777, opened 2020-12-20
CI: Use a single workflow for generating packages.json

thomashoneyman:
As discussed in #776 with @f-f, the master branch is a protected branch, which prevents us from committing to it via a GitHub Action. This PR keeps those protections in place, but it removes the 'tests' workflow from being a required check. That way we can commit a new packages.json file to pull requests and still merge them, even though tests won't re-run on the generated file.


f-f (approved):

thomashoneyman:
🎵 nightmare 🎵 https://github.com/purescript/package-sets/pull/778

This is *still* not working correctly on forks, unfortunately. I've spent more time digging around to understand why, and I came across a new event that GitHub added in August which seems like the actual solution for our problems:

https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#pull_request_target

This event appears to be designed exactly for our use case. Triggering a workflow from _this_ event allows the workflow to be run in the context of the base branch, and for us to pass through repository secrets (like the token used to make the Git commit). That means that our checkout should finally work correctly and that we can go back to protecting the branch with a required 'tests' check.

PR #778, opened 2020-12-21
Add substitute and return

ursi:


thomashoneyman:
@ursi Do you mind if I hijack this PR to test things out a little bit, as it's a failing case? I'll push some tweaks to the workflow directly here.

ursi:
@thomashoneyman yeah go for it

ursi:
@thomashoneyman so what's the current status of including/not including an updated `packages.json`?

ursi:
also would you mind building a new release :) thanks

thomashoneyman:
@ursi Unfortunately, we're reverting to requiring that you run `nix-shell --run 'make generate'` on your pull requests for the time being. It's a bit of a rabbit hole trying to get the commit + workflow run + reported checks to all play well together. If I'm able to come back to it amidst 0.14 updates I'd still like to.

I've already done a release today, so I'll do the next release tomorrow. Thanks!

thomashoneyman:
https://github.com/purescript/package-sets/releases/tag/psc-0.13.8-20201223

ursi:
thank you. Is there any reason to not automate this process? e.g. just have a new release build every day a PR was merged?

thomashoneyman:
We want to, but iirc the last time @f-f explored that there were reasons not to. We've both been super busy with the 0.14 release and haven't been able to dedicate much time to package sets & the registry in the meantime, but once 0.14 is out we'd like to polish the package sets experience with things like automated releases.

f-f:
The reason why this is hard to automate is that it gets quite tricky to get the automation right between compiler versions, especially breaking ones, because we have to do a few steps to update to a new compiler version, like updating the release script, the CI, etc. We could have a checklist, but the price of getting things wrong is higher than the annoyance of cutting a new tag, especially given that from the maintainers' side is just about running the release script once.

PR #779, opened 2020-12-22
Add undefined-is-not-a-problem

paluh:


paluh:
What can I say.. thanks Thomas!

PR #781, opened 2020-12-23
CI: Add update-packages workflow

thomashoneyman:


thomashoneyman:
(The workflow_run event [only runs if the workflow is committed to the base branch](https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#workflow_run), so I'll go ahead and merge this for testing purposes).

PR #784, opened 2020-12-23
Add collegevine/elmish

fsoikin:
1. I hope I did good. I tried to follow the various instructions and guidelines, but this transitional state is mighty confusing :-/
2. There is no corresponding PR to `new-packages.json` in the registry repo, because this package is already in `bower-packages.json`. It is my understanding that the two shouldn't overlap.
3. There will also be `elmish-html`, but it will have to be in a separate PR, because it depends on `elmish`, so I can't give it a proper `spago.dhall` file until `elmish` is in the set. Unless I'm missing something?

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -1035,6 +1035,25 @@
     "repo": "https://github.com/purescript/purescript-either.git",
     "version": "v4.1.1"
   },
+  "elmish": {
+    "dependencies": [
+      "aff",
+      "argonaut-core",
+      "console",
+      "debug",
+      "either",
+      "foreign-object",
+      "functions",
+      "maybe",
+      "prelude",
+      "record",
+      "tuples",
+      "typelevel-prelude",
+      "web-html"
+    ],
+    "repo": "https://github.com/collegevine/purescript-elmish.git",
+    "version": "v0.3.0"
+  },
   "email-validate": {
     "dependencies": [
       "aff",
--- a/src/groups/collegevine.dhall
+++ b/src/groups/collegevine.dhall
@@ -1,0 +1,22 @@
+{ elmish =
+    { dependencies =
+        [ "aff"
+        , "argonaut-core"
+        , "console"
+        , "debug"
+        , "either"
+        , "foreign-object"
+        , "functions"
+        , "maybe"
+        , "prelude"
+        , "record"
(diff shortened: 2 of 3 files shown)


thomashoneyman:
This looks good! Just waiting on CI to pass. You're right that you won't be able to add `elmish-html` without `elmish`, but you can add them both together if you'd like (you'd just add it alongside `elmish` in the collegevine.dhall file).

You can do that in this PR if you'd like, in which case I'll keep it open, or I can go ahead and merge and you can do that in a subsequent PR. Up to you -- just let me know.

fsoikin:
Thank you @thomashoneyman, I'd prefer a subsequent PR. I already have everything lined up.

fsoikin:
@thomashoneyman I suppose I need to ask somebody (you?) to create a new package set release, right? Or can I somehow proceed with `elmish-html` without creating a separate package set release? If so, what would I put in `packages.dhall`?

thomashoneyman:
You can go ahead and add elmish-html to this repository. Add it the same way you added elmish. Once that’s done, I’ll make today’s release. Once the release is up then you can freely install the two libraries in your local projects.

fsoikin:
But if I add `elmish-html`, it won't compile, because `elmish` is not in the set yet, so the verification phase will fail. Or am I missing something?

thomashoneyman:
Sorry about the confusing process!

The package will compile because all of its dependencies exist in the package set in the pull request. That’s why you can add multiple packages which depend on each other at the same time. Or, in this case, since elmish is in the master branch that you’ll be merging into, you’re also good to go.

A package set is really just a collection of Dhall files and the resulting packages.json file. A package set release is a snapshot of that JSON file at a moment in time which Spago or psc-package knows how to download and use.

Does this clarify things at all?

f-f:
@fsoikin the CI uses the local package set to run `verify`, not the one from the package's repo. This is also the reason why you could have added both packages in this PR. Makes sense?

fsoikin:
Ohhhh, I see, it finally clicked. The package's local `packages.dhall` doesn't matter for adding to the set, so I can extra-dep `elmish` for now and then modify after the package set is out, but without cutting a new version.

PR #792, opened 2021-01-11
Add EpicOrange/react-keybind

EpicOrange:
Hey, this is my first package! And I have a question about this process.

Right now `react-keybind` has two dependencies, `react-basic` and `react-basic-classic`. The second one, `react-basic-classic`, is published in package-sets, but not published on [purescript/registry](https://github.com/purescript/registry) or Pursuit. So the only way `bower.json` can find it is via a direct link:

```
    "dependencies": {
        "purescript-react-basic": "^v15.0.0",
        "purescript-react-basic-classic": "git://github.com/lumihq/purescript-react-basic-classic.git#v1.0.1"
    }
```

This is what's currently published at `react-keybind`. I understand that directly linking to the dependency repo url might lead to some reproducibility snags if the dependency decides to move its url (pretty unlikely) so is this an okay configuration?

In other news, everything seems to verify okay:

```
[info] Verifying 1 packages, this might take a while..
[info] Verifying package "react-keybind"
[info] Build succeeded.
[info] Successfully verified "react-keybind"
```

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -3594,6 +3594,14 @@
     "repo": "https://github.com/robertdp/purescript-react-halo.git",
     "version": "v0.2.3"
   },
+  "react-keybind": {
+    "dependencies": [
+      "react-basic",
+      "react-basic-classic"
+    ],
+    "repo": "https://github.com/EpicOrange/purescript-react-keybind.git",
+    "version": "v0.8.1"
+  },
   "react-queue": {
     "dependencies": [
       "exceptions",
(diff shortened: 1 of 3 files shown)


f-f (approved):

f-f:
@EpicOrange thank you! About your question: `react-basic-classic` [is on the registry](https://github.com/purescript/registry/blob/4641a3bcf91cb4f8886ee021991ba534048a4ebf/new-packages.json#L23), but not on the _Bower_ registry, which is why you have to link it like that in your Bowerfile. I think your setup is fine, and CI has passed, so this was all good to merge :slightly_smiling_face:

PR #797, opened 2021-01-18
Adds graphql-client

roryc89:


f-f:
It looks like CI is failing because the `packages.json` was updated to be empty - could you run `nix-shell --run 'make'` and commit all the files?

Another note: could you revert the change to the `Makefile`?

f-f:
@roryc89 it looks like you should add `graphql-parser` too?

CI fails with:
```
The following packages do not exist in your package set:
   - graphql-parser
```

roryc89:
@f-f 

> @roryc89 it looks like you should add graphql-parser too?

Sorry I'm not sure where I should add it to. Before I read your comment I just tried removing from the dependencies.

f-f:
@roryc89 it's fine if it's not included yet, you can just add it along with your package here in this PR - the definition should go in the `src/groups/meeshkan.dhall` file. Makes sense?

f-f (approved):
This looks great now, thank you!

PR #803, opened 2021-02-02
Add aws sdk basic

hivemind-js:


f-f:
Hi! It looks like CI is failing because you added the package with its `purescript-` prefix, while the convention here in package set is to omit that. Could you remove the prefix?

hivemind-js:
Geez I'm blind. Thx for the hint!

f-f:
You're welcome :slightly_smiling_face:

f-f (approved):

Issue #804, opened 2021-02-02
how can I try packages that are already updated to 0.14 with spago


srghma:
seems like I have to:
- download this repo
- change version to "master" in files, e.g. in https://github.com/purescript/package-sets/blob/master/src/groups/purescript.dhall
- change

```
let upstream =
      https://github.com/purescript/package-sets/releases/download/psc-0.13.6-20200309/packages.dhall sha256:9221987b4e7ea99ccd0efbe056f7bebc872cd92e0058efe5baa181d73359e7b3
```

to 

````
let upstream =
      ~/packages/xxxx/src/packages.dhall
```

srghma:
there is also https://github.com/purescript/package-sets/blob/prepare-0.14/src/groups/purescript.dhall

but it seems like outdated

am I missing some other way?

JordanMartinez:
You should be using this file: https://github.com/purescript/package-sets/blob/prepare-0.14/src/packages.dhall

For example:
```dhall
{- spago.dhall -}
{ name = "library"
, packages = https://github.com/purescript/package-sets/blob/prepare-0.14/src/packages.dhall
, dependencies = 
    [ "prelude"
    , "effect"
    ]
, sources = [ "src/**/*/.purs" ]
}
```
The libraries that currently compile are tracked here: https://github.com/purescript/package-sets/blob/prepare-0.14/src/updatedLibs.dhall#L8-L137 However, this PR (https://github.com/purescript/package-sets/pull/802) failed to build, so there might be some issues with the `prepare-0.14` branch right now.

Lastly, we're going to make some breaking changes in https://github.com/purescript/purescript-profunctor/issues/23

We make no guarantees about this branch being "stable" as we may change things unexpectedly at any time. So, do so at your own risk.

thomashoneyman:
The PureScript 0.14 package set has been released, so you can use 0.14 packages like any other set.

PR #809, opened 2021-02-08
Prepare 0.14

srghma:
https://github.com/purescript/package-sets/issues/804

@JordanMartinez

used with https://github.com/srghma/my-purescript-package-sets/blob/master/mypackages.dhall

__________

also I have made a command to update packages

```
sd 'Data.Generic.Rep.Show' 'Data.Show.Generic' $(fd --no-ignore --hidden --type f .purs .)
sd 'Data.Generic.Rep.Eq' 'Data.Eq.Generic' $(fd --no-ignore --hidden --type f .purs .)
sd 'Data.Generic.Rep.Ord' 'Data.Ord.Generic' $(fd --no-ignore --hidden --type f .purs .)
sd 'Data.Argonaut.Decode.Generic.Rep' 'Data.Argonaut.Decode.Generic' $(fd --no-ignore --hidden --type f .purs .)
sd 'Data.Argonaut.Encode.Generic.Rep' 'Data.Argonaut.Encode.Generic' $(fd --no-ignore --hidden --type f .purs .)
sd 'Data.Char.Unicode' 'Data.String.Unicode' $(fd --no-ignore --hidden --type f .purs .)
sd 'Data.Bifunctor.Wrap' 'Data.Functor.Wrap' $(fd --no-ignore --hidden --type f .purs .)
sd '\# Type' 'Row Type' $(fd --no-ignore --hidden --type f .purs .)
sd 'RowList\) ' 'RowList Type) ' $(fd --no-ignore --hidden --type f .purs .)
sd ', kind ([A-Z])' ', $1' $(fd --no-ignore --hidden --type f .purs .)
sd '\( kind ([A-Z])' '( $1' $(fd --no-ignore --hidden --type f .purs .)
```

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -117,7 +117,6 @@
       "arrays",
       "effect",
       "foreign-object",
-      "generics-rep",
       "identity",
       "integers",
       "maybe",
@@ -149,7 +148,6 @@
     "dependencies": [
       "argonaut-codecs",
       "argonaut-core",
-      "generics-rep",
       "record"
     ],
     "repo": "https://github.com/purescript-contrib/purescript-argonaut-generic.git",
@@ -169,7 +167,6 @@
       "arrays",
       "control",
       "foldable-traversable",
-      "generics-rep",
       "maybe",
       "nonempty",
       "prelude",
@@ -414,7 +411,6 @@
       "either",
       "foldable-traversable",
       "foreign-object",
-      "generics-rep",
       "maybe",
       "newtype",
       "prelude",
@@ -530,7 +526,6 @@
   },
   "byte-codec": {
     "dependencies": [
-      "generics-rep",
       "integers",
       "sized-vectors"
     ],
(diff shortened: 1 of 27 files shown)


JordanMartinez:
Thanks for putting in some time to update the package set. However, there are other changes that need to be made and we aren't going to update the dependencies for packages outside of core as they aren't ours. (I did replace some packages' `globals` dependency with `numbers` previously, but I realize now that I should not have done that.) We also have plans to run a script that would likely overwrite the changes here even if I did merge. So, for these reasons, I'm going to close this.

However, if you want to submit a PR fixing the dependencies for the `node-http` and `web-xhr` packages in a separate PR, I'm willing to merge that.

Before submitting a PR like this in the future, could you discuss this first in an issue? I wasn't sure why you linked to #804 initially and still don't entirely know.

srghma:
> I wasn't sure why you linked to #804 initially and still don't entirely know.

To show what deps I am using

PR #820, opened 2021-03-02
Update purescript-variant to v0.14.0

JordanMartinez:


JordanMartinez:
Let me know if there's anything else I need to do here.

thomashoneyman:
I wonder if there’s some non-exhausting way to see what downstream libraries this unblocks. Maybe just a grep and an eyeball over dependencies to see? It would be nice if we could uncomment some libraries each time we merge a cornerstone library like this one.

thomashoneyman (approved):

f-f:
@thomashoneyman no worries, I'll do a pass of uncommenting unblocked libraries in a couple of days or so. I might patch spago to help us here in the process.

JordanMartinez:
If you clone and run the [`rebuildRemainingPackages.sh`](https://github.com/JordanMartinez/updatePsLibs/blob/master/rebuildRemainingPackages.sh), you can then check the [`libDeps.txt` file](https://github.com/JordanMartinez/updatePsLibs/blob/master/libDeps.txt) to see what's been unblocked.

Not 100% accurate as some packages added dependencies since then, but it's still a better guess than what we currently have.

PR #823, opened 2021-03-02
Add turf package

jisantuc:


src/groups/jisantuc.dhall
```diff
@@ -0,0 +1,15 @@
+{ turf =
+  { dependencies =
+    [ "argonaut-codecs"
+    , "argonaut-core"
+    , "assert"
+    , "console"
+    , "effect"
+    , "foreign-object"
+    , "psci-support"
+    , "quickcheck"
```
thomashoneyman, line 10: Are all of these used in the source code? You don't have to include test dependencies (`assert`, `console`, and `psci-support` (for the repl) stand out).

thomashoneyman (approved):

src/groups/jisantuc.dhall
jisantuc: `assert`, `psci-support`, and probably `console` even at this point are not. `quickcheck` is for the arbitrary instances, so I think that counts. I'll remove them and see if `make` is still happy.

jisantuc: You're right, a number of those are unused. I assumed I needed everything declared in my `spago.dhall` in the package

`spago verify` is so helpful 😎

PR #839, opened 2021-03-05
Update httpure

sigma-andex:


cprussin (approved):

cprussin:
thanks @sigma-andex

src/groups/cprussin.dhall
```diff
@@ -53,7 +53,43 @@
...
+    , "refs"
+    , "spec"
+    , "strings"
+    , "tuples"
+    , "type-equality"
+    , "unsafe-coerce"
+    ]
+  , repo = "https://github.com/cprussin/purescript-httpure.git"
+  , version = "v0.11.0"
+  }
```
thomashoneyman, line 91: Could you remove the commented-out definition for `httpure` above?

Also, while the previous version already had it, it looks like there are some unnecessary dependencies in here. We only need production dependencies in package sets. `psci-support` and `spec` could most likely be dropped, and maybe even `console` if it isn't imported anywhere.

sigma-andex: Yeah will have a look at it. I think there is also still something broken. Will have a look at it tomorrow

thomashoneyman: Looks like you're missing the `js-uri` dependency

sigma-andex:
Ok I think I finally got it. Thanks a lot for the explanations @cprussin  and @thomashoneyman  :-)

PR #860, opened 2021-03-14
Update `b64` to v0.0.7

menelaos:


hivemind-js:
@menelaos I think this just needs a rebase

PR #863, opened 2021-03-15
Update colors, halogen, and functors, and add css

thomashoneyman:


thomashoneyman:
@f-f I've adjusted some package dependencies to remove `debug`, as this seems like it would almost certainly be a mistake to include. Alternately we could drop `debug` for now and only update it once the downstream packages move it to their development dependencies.

thomashoneyman:
Well, some packages _do_ use `debug` internally, so I'm not going to update it for now.

PR #864, opened 2021-03-15
Upgrade debug, elmish, and elmish-html

fsoikin:
The new v0.5.1 of `elmish` is compatible with the new v5.0.0 of `debug`

~There is a bit of a confusion about where `debug` should be in `devDependencies` in `bower.json` or not (see https://github.com/collegevine/purescript-elmish/pull/34#issuecomment-799673449), but this shouldn't affect package sets as far as I understand.~ Resolved now

**Edit**: ok, it looks like there is some other incompatibility in some other library

**Edit 2**: aha, found it. `debuggest` [has the same module name](https://github.com/ursi/purescript-debuggest/blob/master/src/Debug.purs#L1).

**Edit 3**: I commented out `debuggest` for now. Not sure if that's acceptable.

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -812,7 +812,7 @@
       "web-html"
     ],
     "repo": "https://github.com/collegevine/purescript-elmish.git",
-    "version": "v0.4.0"
+    "version": "v0.5.0"
   },
   "elmish-html": {
     "dependencies": [
@@ -820,7 +820,7 @@
       "foreign-object"
     ],
     "repo": "https://github.com/collegevine/purescript-elmish-html.git",
-    "version": "v0.2.0"
+    "version": "v0.3.0"
   },
   "email-validate": {
     "dependencies": [
--- a/packages.json
+++ b/packages.json
@@ -693,14 +693,6 @@
     "repo": "https://github.com/garyb/purescript-debug.git",
     "version": "v5.0.0"
   },
-  "debuggest": {
-    "dependencies": [
-      "prelude",
-      "unsafe-coerce"
-    ],
-    "repo": "https://github.com/ursi/purescript-debuggest.git",
-    "version": "v0.4.0"
-  },
   "decimals": {
     "dependencies": [
       "maybe"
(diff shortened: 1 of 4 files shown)


thomashoneyman:
Oof, that's a little unfortunate. I think it's sensible that `Debug` is reserved by the `debug` library, especially given its canonical place in the PureScript ecosystem, but I'm a little reluctant to drop `debuggest` without speaking with @ursi first.

@ursi -- I'm sorry to introduce a hassle, but would you be open to renaming your module to `Debuggest` to match the package name? We didn't realize there would be a conflict with the new `debug` module name.

ursi:
I took `Debug` because no one had bothered to put `debug` in the old package set, so I figured it wasn't very important. Of course, I would like to keep the module name, but if `debug` is making a return and it's important to the ecosystem, then I guess I gotta step aside.

fsoikin:
So would you like `debuggest` to be removed from the set, or would you like a pull request changing the module name instead @ursi ?

ursi:
@fsoikin you can remove it or leave it commented. I'll figure at what to name it at some point and put in a PR.

thomashoneyman:
`debug` is rarely depended on by libraries (except as a dev dependency which wouldn't be tracked in package sets), but it's the standard for debugging in applications. Applications tend to lag the package set so `debug` was slow to get in for 0.14, but it's a commonly-used package.

ursi:
> Applications tend to lag the package set so debug was slow to get in for 0.14, but it's a commonly-used package.

Just so we're on the same page, I didn't snipe the name in the 0.14 package set. I haven't moved any of my packages to 0.14 yet. I had this in the packages set for 0.13, and I guess it was moved over automatically.

thomashoneyman:
I didn’t mean to insinuate that you did; in fact, the module name for debug only changed with the 0.14 release and started to conflict. I’m sorry about the hassle. I appreciate that you’re willing to consider changing the module name.

PR #867, opened 2021-03-16
Update aws-sdk-basic

tiphdousset:


src/groups/hivemindtechnologies.dhall
```diff
@@ -1,4 +1,19 @@
...
+{ aws-encryption-sdk =
+  { dependencies =
+    [ "aff-promise"
+    , "console"
+    , "debug"
+    , "effect"
+    , "node-buffer"
+    , "psci-support"
+    , "spec"
+    , "spec-discovery"
```
thomashoneyman, line 10: It's not a problem, and I can go ahead and merge this, but are these dependencies all necessary for production users? For example, `console`, `debug`, `psci-support`, `spec`, and `spec-discovery` are development dependencies for most libraries. If you want to merge as-is then just let me know and I can get this in. Thanks!

tiphdousset: Good point, thanks for the info. I'll have a closer look. 
The dependencies you mention are related to another library than the one currently updated, so yes please merge this PR for now. 
The changes you mention pop up due to a re-ordering in the group file.
Thanks again for the hint!!

PR #880, opened 2021-03-19
Add yaml-next

ad-si:


thomashoneyman:
@ad-si could you also run `nix-shell --run make` locally and commit the resulting packages.json file? Thanks!

src/groups/ad-si.dhall
```diff
@@ -3,4 +3,20 @@
...
+  { dependencies =
+    [ "argonaut-codecs"
+    , "argonaut-core"
+    , "console"
+    , "effect"
+    , "foreign"
+    , "functions"
+    , "ordered-collections"
+    , "psci-support"
+    , "spec"
```
thomashoneyman, line 16: These look like development dependencies -- while it's fine to add them, you only have to include dependencies used in the source.

ad-si:
@thomashoneyman Thanks for the infos! Commit the changes to the `package.json` file and remove the dev dependencies.

PR #883, opened 2021-03-20
Add cirru-parser

jiyinyiyong:
Corresponded PR in https://github.com/purescript/registry/pull/147/files . First time to submit package here...

I got a some confusions:

1. in `dependencies` field, I listed the packages I used manually. however among them the packages used in tests are not listed. when I was in Node.js there are `dependencies` and `devDependencies`. is that okay I leave out the `devDependencies` part in this?
2. does it also mean I need to PR every time I bumped version since `version` is locked here, even though my package is still in early stage? I tried but it appears I can't run `pulp publish` without the depended package listed in package-sets, I Googled many times and didn't see a way of maintaining dependencies locally(given that my package still in early stage and will probably change).

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -541,6 +541,15 @@
     "repo": "https://github.com/icyrockcom/purescript-cheerio.git",
     "version": "v0.2.3"
   },
+  "cirru-parser": {
+    "dependencies": [
+      "arrays",
+      "maybe",
+      "prelude"
+    ],
+    "repo": "https://github.com/Cirru/parser.purs.git",
+    "version": "v0.0.3"
+  },
   "clipboardy": {
     "dependencies": [
       "aff",
--- a/src/groups/cirru.dhall
+++ b/src/groups/cirru.dhall
@@ -1,0 +1,10 @@
+{ cirru-parser =
+  { dependencies =
+    [ "arrays"
+    , "maybe"
+    , "prelude"
+    ]
+  , repo = "https://github.com/Cirru/parser.purs.git"
+  , version = "v0.0.3"
+  }
+}
--- a/src/packages.dhall
+++ b/src/packages.dhall
@@ -16,6 +16,7 @@ let packages =
       ⫽ ./groups/cdepillabout.dhall
       ⫽ ./groups/chekoopa.dhall
       ⫽ ./groups/citizennet.dhall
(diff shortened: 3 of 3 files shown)


thomashoneyman:
>     1. in `dependencies` field, I listed the packages I used manually. however among them the packages used in tests are not listed. when I was in Node.js there are `dependencies` and `devDependencies`. is that okay I leave out the `devDependencies` part in this?

Yes, you should leave out dev dependencies and only include dependencies used in your source.

>     2. does it also mean I need to PR every time I bumped version since `version` is locked here, even though my package is still in early stage? I tried but it appears I can't run `pulp publish` without the depended package listed in package-sets, I Googled many times and didn't see a way of maintaining dependencies locally(given that my package still in early stage and will probably change).

Yes, the `version` is locked here. You can either PR each time you bump the version, or you can wait for us to pick up the change (we update versions every couple days and check for updates automatically).

You don't have to have your library and its dependencies in the package set in order to work on it. You can rely on your packages locally -- see [the section of the Spago documentation about overriding the package set to a local package](https://github.com/purescript/spago#override-a-package-in-the-package-set-with-a-local-one).

Note that `pulp` doesn't know anything about package sets; `pulp publish` is used to push your package documentation to Pursuit and it uses your local `bower.json` file to do so, but it's unrelated to package sets.

There's one last thing -- if you are failing to run `pulp publish` because you have a dependency listed in your `bower.json` (like `cirru-parser`) and Bower says it doesn't know about the package, then that means that you published the package after the Bower registry officially closed. In that case, just make sure to provide the full Git URL to Bower instead of just the package name. For example:

```diff
- "cirru-parser": "^1.0.0"
+ "cirru-parser": "https://github.com/jiyinyiyong/purescript-cirru-parser.git#^1.0.0"
```

jiyinyiyong:
thanks a a lot.

```diff
- "cirru-parser": "^1.0.0"
+ "cirru-parser": "https://github.com/jiyinyiyong/purescript-cirru-parser.git#^1.0.0"
```

this looks a lot like what I saw in Nim packages 😄

thomashoneyman (approved):

jiyinyiyong:
it has been 3+ days, could you tag a new release?

thomashoneyman:
`psc-0.14.0-20210324`

PR #901, opened 2021-04-01
Add fuzzy, halogen-select, and halogen-formless

thomashoneyman:
cc @davezuch


davezuch:
Thanks @thomashoneyman!

PR #923, opened 2021-04-19
Add a v0.14.1 compatible package set by using current one

JordanMartinez:
Temporarily addresses the issue raised in purescript/spago#777


f-f:
The real fix is even simpler: one needs the change in https://github.com/purescript/package-sets/commit/e3b1ce0d90b7941474d873785d54c3560bcbedf3 and then to run the `./release.sh` script

PR #927, opened 2021-04-27
Updates

f-f:


f-f:
@ursi I tried upgrading `mason-prelude` to the latest version - however it looks like it now depends on `debuggest`, which is not in the set because of the conflicting `Debug` module name.
I'd like to fix it up and keep it in the set, but from https://github.com/ursi/purescript-debuggest/issues/1 it looks like you might not be interested in doing this change.
So in this PR I'm dropping `mason-prelude` and all its dependents from the set. Is that fine by you or should we go for some other solution?

ursi:
yep, that's totally fine! Those two dependants have actually already been refactored to no longer depend on `mason-prelude`, I just hadn't gotten around to updating them here.

f-f:
@ursi great, thank you!

PR #940, opened 2021-05-17
Add float32

jamesdbrock:
I'm upgrading this package to purs 0.14 and adding it back to package-sets on behalf of @athanclark .

https://github.com/athanclark/purescript-float32/pull/3

Is that cool? What's the etiquette here?


f-f:
@jamesdbrock the package you added as `float32` is pointing to a different URL than what the `float32` entry in the registry points to, and as a policy we don't allow this here in package-sets (this is the whole point of having a "registry").
We should really have a CI check for this, I'll look into adding one.
If you'd like to move forward with this the recommended way would be to fork the package under a different name, and add it to the registry and here with the new name.

f-f:
> We should really have a CI check for this, I'll look into adding one

Added a check for this in #941, CI is now failing:

jamesdbrock:
Oh yeah right, the registry, I forgot. Hm.

PR #949, opened 2021-05-31
Added typelevel-codec-json

JordanMartinez:
Pending purescript/registry#179

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -4075,6 +4075,33 @@
     "repo": "https://github.com/bodil/purescript-typelevel.git",
     "version": "v6.0.0"
   },
+  "typelevel-codec-json": {
+    "dependencies": [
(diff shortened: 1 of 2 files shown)


thomashoneyman:
@f-f do we have restrictions on case sensitivity?

JordanMartinez:
Ah... I think this is the second time this has happened. So, I think we do.

f-f:
@thomashoneyman @JordanMartinez I introduced a check recently, in https://github.com/purescript/package-sets/pull/941, where we verify that the URL here in package-set matches the one in the registry files. It can be made case insensitive if need be

JordanMartinez:
This library didn't do what I thought it would, so I'm actually glad it wasn't added to the registry.

PR #954, opened 2021-06-12
Update two-or-more

i-am-the-slime:
Also try to update PureScript to 0.14.2


i-am-the-slime:
Not sure what needs to be done here because this change removes instance names in the library so it's only compatible with 0.14.2.

f-f:
I think we can just cut a new set for 0.14.2

f-f (approved):
This is good, thank you! :clap:

PR #970, opened 2021-07-13
Misc. updates

thomashoneyman:
Updates to various packages. @f-f this includes the metadata bump since 0.14.3 is out; is there anything special that needs to be done wrt this release? Should that be separated out?


thomashoneyman:
I've removed the metadata bump for now since it also needs a PureScript bump in CI.

f-f:
Yeah, in addition to updating the `metadata` package we also have to (1) bump the compiler in CI and (2) update the release script - e.g. see https://github.com/purescript/package-sets/pull/955

PR #710, opened 2020-09-26
Add @maxdeviant's packages

maxdeviant:
This PR adds the following packages of mine to the package set:

- [heckin](https://pursuit.purescript.org/packages/purescript-heckin/1.0.3)
- [npm-package-json](https://pursuit.purescript.org/packages/purescript-npm-package-json/1.2.0)
- [which](https://pursuit.purescript.org/packages/purescript-which/0.1.0)

PR #977, opened 2021-07-30
Update `heckin` and `npm-package-json` to PureScript 0.14

maxdeviant:
This PR updates the following packages to work with PureScript 0.14 and the latest package set:

- `heckin`
- `npm-package-json`

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -1775,6 +1775,24 @@
     "repo": "https://github.com/purescript-halogen/purescript-halogen-vdom.git",
     "version": "v7.0.1"
   },
+  "heckin": {
+    "dependencies": [
+      "aff",
+      "arrays",
+      "effect",
+      "foldable-traversable",
+      "maybe",
+      "prelude",
+      "psci-support",
+      "spec",
+      "strings",
+      "transformers",
+      "tuples",
+      "unicode"
+    ],
+    "repo": "https://github.com/maxdeviant/purescript-heckin.git",
+    "version": "v2.0.0"
+  },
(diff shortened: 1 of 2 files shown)


packages.json
```diff
@@ -1775,6 +1775,24 @@
...
+  "heckin": {
+    "dependencies": [
+      "aff",
+      "arrays",
+      "effect",
+      "foldable-traversable",
+      "maybe",
+      "prelude",
+      "psci-support",
+      "spec",
```
thomashoneyman, line 1787: These look like they might be dev dependencies and could potentially be omitted

packages.json
```diff
@@ -2549,6 +2567,20 @@
...
+  "npm-package-json": {
+    "dependencies": [
+      "argonaut",
+      "control",
+      "either",
+      "foreign-object",
+      "maybe",
+      "ordered-collections",
+      "prelude",
+      "psci-support"
```
thomashoneyman, line 2579: It's not necessary to include `psci-support` here in the package sets, but you can if you'd like!

thomashoneyman (approved):
Approving, but leaving it open for a day in case you wanted to look at the comments I left. thanks!

maxdeviant:
> Approving, but leaving it open for a day in case you wanted to look at the comments I left. thanks!

Ah, yes, I now recall having this conversion [before](https://github.com/purescript/package-sets/pull/710) 😅 

Very much looking forward to the new registry (as I'm sure we all are)!

PR #986, opened 2021-08-12
Re-add quantities

mhmdanas:


src/groups/sharkdp.dhall
```diff
@@ -15,14 +15,27 @@
...
+    , "foldable-traversable"
+    , "lists"
+    , "math"
+    , "maybe"
+    , "newtype"
+    , "nonempty"
+    , "numbers"
+    , "pairs"
+    , "prelude"
+    , "psci-support"
```
mhmdanas, line 32: Do I remove this? It's not really used in the code.

thomashoneyman: Yep! No need to include it.

PR #991, opened 2021-08-22
adds purescript-phaser

lfarroco:
https://github.com/lfarroco/purescript-phaser

### DIFF
--- a/src/groups/lfarroco.dhall
+++ b/src/groups/lfarroco.dhall
@@ -1,0 +1,6 @@
+{ purescript-phaser =
+  { dependencies = [ "aff", "effect", "prelude", "functions" ]
+  , repo = "https://github.com/lfarroco/purescript-phaser.git"
+  , version = "v0.0.1"
+  }
+}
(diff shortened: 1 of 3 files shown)


src/groups/lfarroco.dhall
```diff
@@ -0,0 +1,6 @@
+{ purescript-phaser =
```
thomashoneyman, line 1: In the package sets we don't use the `purescript-` prefix. Would you mind updating the name of the library in your `.dhall` file to be this?

```suggestion
{ phaser =
```

thomashoneyman (commented):
Once you update the package name I can go ahead and get this merged. Sorry for the hassle!

src/groups/lfarroco.dhall
lfarroco: thanks for the guidance, will do!

lfarroco:
Should I rename it on https://github.com/purescript/package-sets/pull/991/files#diff-c508188958b68ce78976e33b66f47762f45f4a7bc134c5781cc38ee9ee81589aR3287 too? It is also necessary to create a new entry on `purescript/registry`?

thomashoneyman:
Oh, sorry, yes. It should be renamed in both places. (Running `nix-shell --command make` will also do it for you.)

PR #1011, opened 2021-11-01
update purescript-phaser

lfarroco:


f-f:
It looks like CI is failing because the `packages.json` was not updated - could you run `nix-shell --run 'make'` and commit all the files?

PR #1013, opened 2021-11-07
Update `elmish` to v0.5.6

maxdeviant:
This PR updates `elmish` to v0.5.6.

https://github.com/collegevine/purescript-elmish/compare/v0.5.1...v0.5.6


src/groups/collegevine.dhall
```diff
@@ -2,23 +2,34 @@
...
+    , "web-dom"
     , "web-html"
     ]
   , repo = "https://github.com/collegevine/purescript-elmish.git"
-  , version = "v0.5.1"
+  , version = "v0.5.6"
   }
 , elmish-html =
-  { dependencies = [ "elmish", "foreign-object" ]
+  { dependencies = [ "elmish", "foreign-object", "record" ]
```
maxdeviant, line 32: `elmish` previously depended on `record`, but does not anymore.

However, `elmish-html` still depends on `record`, but it is not listed as one of its direct dependencies. Previously it was relying on `record` transitively through `elmish`.

Adding it as a dependency here to correct the issue.

PR #1025, opened 2021-11-10
Update `aff-bus` to v5.0.1

maxdeviant:
This PR updates `aff-bus` to v5.0.1.

https://github.com/purescript-contrib/purescript-aff-bus/compare/v5.0.0...v5.0.1

### DIFF
--- a/src/groups/purescript-contrib.dhall
+++ b/src/groups/purescript-contrib.dhall
@@ -27,7 +27,7 @@
 , aff-bus =
   { dependencies = [ "avar", "effect", "prelude" ]
   , repo = "https://github.com/purescript-contrib/purescript-aff-bus.git"
-  , version = "v5.0.0"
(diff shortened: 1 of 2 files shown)


src/groups/purescript-contrib.dhall
```diff
@@ -25,7 +25,22 @@
...
+    [ "aff"
+    , "avar"
+    , "console"
+    , "effect"
+    , "either"
+    , "exceptions"
+    , "foldable-traversable"
+    , "lists"
+    , "prelude"
+    , "psci-support"
```
JordanMartinez, line 38: This dependency isn't needed.

maxdeviant: Removed!

JordanMartinez:
I've cancelled the preceding CI builds, so that only the latest commit will be built instead.

PR #1031, opened 2021-11-17
Update `unicode` to v5.0.1

maxdeviant:
This PR updates `unicode` to v5.0.1.

https://github.com/purescript-contrib/purescript-unicode/compare/v5.0.0...v5.0.1

### DIFF
--- a/src/groups/purescript-contrib.dhall
+++ b/src/groups/purescript-contrib.dhall
@@ -474,23 +474,7 @@
   , version = "v6.0.3"
   }
 , unicode =
-  { dependencies =
-    [ "arrays"
-    , "assert"
-    , "console"
-    , "effect"
-    , "enums"
-    , "foldable-traversable"
-    , "integers"
-    , "maybe"
-    , "partial"
-    , "prelude"
-    , "quickcheck"
-    , "random"
-    , "strings"
-    , "transformers"
-    , "unsafe-coerce"
-    ]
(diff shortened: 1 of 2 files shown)


src/groups/purescript-contrib.dhall
```diff
@@ -474,9 +474,25 @@
...
+    , "assert"
+    , "console"
+    , "effect"
+    , "enums"
+    , "foldable-traversable"
+    , "integers"
+    , "maybe"
+    , "partial"
+    , "prelude"
+    , "quickcheck"
```
maxdeviant, line 488: There were quite a few dependencies added between v5.0.0 and v5.0.1.

Not sure which ones we should be including here (e.g., do we include `quickcheck` even if it's just a test dependency?)

thomashoneyman: We only need to include dependencies in the source code, so nothing that's only a test dependency. I'm assuming that the new dependencies are added just because they were previously transitive?

thomashoneyman: Yea, https://github.com/purescript-contrib/purescript-unicode/pull/35 looks like it was dealing with transitive deps. But we don't need any test-only dependencies included in the package sets.

maxdeviant: Alright, reverted the changes to the dependency list 👍🏻 

I can't wait for Spago targets 😭

PR #1038, opened 2021-11-21
license added

Ebmtranceboy:


maxdeviant:
Hey @Ebmtranceboy, it looks like CI is failing because you need to apply the changes here as well: https://github.com/purescript/package-sets/blob/2d15cde46ad7d2d7a247aa92cf776dacbc63a161/src/groups/ebmtranceboy.dhall#L6-L10

(The `packages.json` is just an artifact we generate from the Dhall files).

Ebmtranceboy (commented):
updating tag in dhall file

maxdeviant:
> updating tag in dhall file

Now it looks like it's failing because a new entry for `geometry-plane` was added to `packages.json` instead of updating the existing entry.

I think if you fix that then CI should pass.

Ebmtranceboy:
I tried to remove the previous entry, then to remove every reference (hoping the generated version would complete). Now, I'm clueless. Maybe generate a new tag ?

f-f:
Hi @Ebmtranceboy! The usual workflow for editing the files here is that contributors edit the `.dhall` files, and then generate the `packages.json` by running `nix-shell --run 'make'`. Could you try that?

Ebmtranceboy:
Hey thanks @f-f  and @maxdeviant . Totally forgotten about nix make ;) Sorry !

Issue #1042, opened 2021-11-25
Hashes for packages in the package sets


cdepillabout:
- Feature Name: Hashes for packages in PureScript package set
- Start Date: 2021-11-25
- RFC PR: #1043

## Summary

It would be nice to have hashes for all the packages in the package set.

This repo distributes package sets where packages are defined like the following:

```dhall
{
  zipperarray =
    { dependencies =
      [ "arrays", "maybe", "prelude", "naturals", "strictlypositiveint" ]
    , repo = "https://github.com/jamieyung/purescript-zipperarray.git"
    , version = "v1.1.0"
    }
}
```

This RFC proposes changing them to also include hashes.  Packages with hashes would look something like this:

```dhall
{
  zipperarray =
    { dependencies =
      [ "arrays", "maybe", "prelude", "naturals", "strictlypositiveint" ]
    , hash = "sha256-fP5dDqsc3/rjZ+WTKFrfIGzuwFnmSQLQg3o+NC17jpY="
    , repo = "https://github.com/jamieyung/purescript-zipperarray.git"
    , version = "v1.1.0"
    }
}
```

Note there is a new `hash` key.

## Motivation

Distributing the package set with hashes for each package would have two nice effects:

- Some build systems have special support for downloading sources that are hashed.  Nix and Bazel both fall into this camp.  The PureScript package set could be directly used by these build systems if each package had a hash.  Currently, an impure step that adds hashes needs to run before the PureScript package set can be used by these systems.

- Support could be added to Spago to check the hash of the packages it is downloading.  Spago could report an error if it downloads a package that doesn't match the specified hash.

## Detailed design

The main change to this `purescript/package-sets` repo would be generating a hash for each package when generating a new package set.  This hash generation code would run right after this line in CI:

https://github.com/purescript/package-sets/blob/c8efa6bd161ba92a4184b58487b628984d8068c9/.github/workflows/release.yml#L18

Here's a PR that implements this proposal: #1043.

Here's an example package set produce with the above PR.  Note that each package contains an additional `hash` field: https://github.com/cdepillabout/package-sets/blob/25b062fa7a5b875aab56c6c45198f2b51ee078fe/packages-with-hashes.dhall

There are a few design questions in this RFC.  I discuss each of them in the following sections, including my suggestions.

### What exactly should be hashed?

[...]

PR #1043, opened 2021-11-25
Add script for adding hashes to each package in the package set

cdepillabout:
This PR adds a script that can add hashes to each package defined in the PureScript package set.  This is implemented as described in the RFC #1042.

Here's an example package set where each package has a hash.  This package set was produced by the `add-hashes.hs` script added in this PR: https://github.com/cdepillabout/package-sets/blob/25b062fa7a5b875aab56c6c45198f2b51ee078fe/packages-with-hashes.dhall

Note that this PR doesn't yet add `add-hashes.hs` to run during the [release CI](https://github.com/purescript/package-sets/tree/79cd8430a8db1b3b82a06207c762203f133adbf9), but that should either be added here, or in a follow-up PR.

----------------------------------------------------------

Here's how you can test this `add-hashes.hs` script locally.

First, generate the package set and confirm that is has been correctly generated:

```console
$ nix-shell --run '>packages.dhall dhall <<< ./src/packages.dhall'
$ head ./packages.dhall
{ abides =
  { dependencies = [ "enums", "foldable-traversable" ]
  , repo = "https://github.com/athanclark/purescript-abides.git"
  , version = "v0.0.1"
  }
, ace =
  { dependencies =
    [ "arrays"
    , "effect"
    , "foreign"
```

Run the `add-hashes.hs` script to add hashes to each package, and confirm the package set with hashes:

```console
$ nix-shell --run add-hashes
$ head packages-with-hashes.dhall
{ abides =
  { dependencies = [ "enums", "foldable-traversable" ]
  , hash = "sha256-nrZiUeIY7ciHtD4+4O5PB5GMJ+ZxAletbtOad/tXPWk="
  , repo = "https://github.com/athanclark/purescript-abides.git"
  , version = "v0.0.1"
  }
, ace =
  { dependencies =
    [ "arrays"
    , "effect"
```

You can see that the `abides` package now contains a hash.

### DIFF
--- a/add-hashes.hs
+++ b/add-hashes.hs
@@ -1,0 +1,144 @@
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+
+import Control.Concurrent (threadDelay)
+import Control.Monad.IO.Class (liftIO)
+import Control.Monad.Trans.Resource (runResourceT)
+import qualified Crypto.Hash.SHA256 as Hash
+import Data.ByteString (ByteString)
+import qualified Data.ByteString.Base64 as Base64
+import Data.Conduit (ConduitT)
+import qualified Data.Conduit
+import qualified Data.Conduit.Binary
+import qualified Data.Either.Validation as Validation
+import Data.Text (Text)
+import qualified Data.Text as Text
+import qualified Data.Text.Encoding
+import Data.Text.Prettyprint.Doc (Pretty)
+import qualified Data.Text.Prettyprint.Doc.Render.Text as Pretty.Text
+import qualified Data.Text.IO
+import Data.Void (Void)
+import qualified Dhall as Dhall
+import qualified Dhall.Core as Dhall
+import qualified Dhall.Map
+import qualified Dhall.Map as Dhall (Map)
+import qualified Dhall.Pretty
+import qualified Dhall.Src as Dhall
+import Network.HTTP.Simple (getResponseBody, getResponseHeaders, getResponseStatus, httpSink, parseRequestThrow)
+import System.IO (hClose, hSetBinaryMode)
+import System.IO.Temp (withSystemTempFile)
+
+main :: IO ()
+main = do
+  text <- Data.Text.IO.readFile "./packages.dhall"
+  dhallExpr <- Dhall.inputExpr text
+  case dhallExpr of
+    Dhall.RecordLit ks -> do
+      let packageList = Dhall.Map.toList ks
+      packageListWithHashes <- traverse getHashForPackage packageList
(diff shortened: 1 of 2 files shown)


default.nix
```diff
@@ -1,7 +1,11 @@
 let
-  pkgs = import (builtins.fetchTarball {
-    url = "https://github.com/NixOS/nixpkgs/archive/21.05.tar.gz";
```
cdepillabout, line 3: In general you don't want to pin to the release tags (like `21.05`), since they are tagged at the beginning of the release cycle.  You instead want to pin to a commit from the channel for a release.  You can find a list of channels and corresponding latest commits here: https://status.nixos.org/

As of right now, the latest commit for the `nixos-21.05` channel is https://github.com/NixOS/nixpkgs/commit/91e6a9a6568f751bb10222a937507e7e151a035e:

image

I put together this PR a few weeks ago, so I am pinning to an older commit from the `nixos-21.05` channel.

default.nix
```diff
@@ -25,6 +29,24 @@ let
...
+          pkgs.haskellPackages.conduit-extra
+          pkgs.haskellPackages.cryptohash-sha256
+          pkgs.haskellPackages.dhall
+          pkgs.haskellPackages.http-conduit
+          pkgs.haskellPackages.prettyprinter
+          pkgs.haskellPackages.resourcet
+          pkgs.haskellPackages.temporary
+        ];
+      }
+      (builtins.readFile ./add-hashes.hs);
```
cdepillabout, line 48: I just used `writeHaskellBin` to define this package, but in a future PR this `add-hashes` script could be expanded to a normal Haskell package and built with either `cabal` directly, or `callCabal2nix` in this `default.nix` file.

add-hashes.hs
```diff
@@ -0,0 +1,144 @@
+{-# LANGUAGE OverloadedStrings #-}
```
cdepillabout, line 1: There are two changes that could potentially be made to this script in a future PR:

- Add error handling for the case when a package source tarball can't be downloaded from GitHub.  (The script could be made to re-try the download a few times before giving up.)

    I ran this script a bunch of times locally while testing it, and I never saw an error.  I imagine the chance that downloading a source tarball from GitHub fails is relatively low.  Although it would still be nice to have reasonable error handling.

- Run the package source tarball fetching and hashing in parallel.

    Even without running in parallel, the downloading-and-hashing process only takes 4 minutes with my slow computer and slow internet connection, so I don't think it is too huge of a problem not to run  it in parallel, at least for now.

I don't think the lack of either of these features should prevent this PR from getting merged in, but they would both be nice to have.

cdepillabout:
Closing this since it seems most work is going into the PureScript Registry.  See https://github.com/purescript/package-sets/issues/1042#issuecomment-981300488.

PR #1059, opened 2022-01-25
update HTTPure to latest version with dependencies

danielstaleiny:
First time doing this. Do I need to validate anything to make sure it is correct ? 

compared dependencies and fixed them in package-sets and up the version. Is there anything eles I need to do to make latest version available ? 

https://github.com/cprussin/purescript-httpure


danielstaleiny:
While patching it locally I was forced to patch httpure-middleware as well. I add that one too, otherwise it does not verify the package.

thomashoneyman:
@danielstaleiny If updating a package will remove downstream packages from the set, we usually try to notify the packages that will be dropped from the set downstream by opening an issue on their repository and give them up to a week to make the changes they need to remain included in the package set. That keeps package set upgrades smooth for users.

I've [opened an issue](https://github.com/joneshf/purescript-httpure-middleware/issues/32) on `httpure-middleware` about this; regardless of the response, we can merge this in a few days.

thomashoneyman:
@danielstaleiny Would you mind removing `httpure-middleware` from the package sets and regenerating the `packages.json` file? Then I can merge this for you!

danielstaleiny:
I removed httpure-middleware both from package.json and from dhall file. However I have no idea how to generate automatically package.json so I did it manually(which I understand is error prone) so for future, what command am I suppose to run to make it re-generate ?

danielstaleiny:
I am blind. I was looking in Readme. 
```
[nix-shell:~/work/github/package-sets]$ ./verify-registry.sh
[SUCCESS] All packages are in the registry
[SUCCESS] All package URLs are matching the Registry
```

danielstaleiny:
No change. 
```sh
λ package-sets master ✓ nix-shell --command make
formatted dhall files
generated to packages.json
```

danielstaleiny:
I get the flow now. and Yes, I just double check to see that there are indeed no changes. (Therefore I did manual change correctly.)

thanks for info and time.

PR #1071, opened 2022-02-25
Upgrade to 0.14.6

thomashoneyman:
Closes #1070

PR #1074, opened 2022-02-28
Make 0.14.7 package set

JordanMartinez:


thomashoneyman:
Can you also update the metadata package and compiler version as done in #1071?

JordanMartinez:
Done! I forgot about that.

PR #1077, opened 2022-03-10
Update string-parsers to v7.0.0

chtenb:
[v7.0.0](https://github.com/purescript-contrib/purescript-string-parsers/releases)


thomashoneyman:
@chtenb It looks like this will cause at least `biscotti-cookie` to be dropped from the set. Usually we notify package authors that their package will be dropped from the set if it isn't updated by opening an issue on their repo stating that a dependency has updated and their package no longer compiles.

Would you mind checking if any other packages will be dropped from the set due to this change?

If you would like, I can take care of opening the notification issues on your behalf. You can also open the issues if you'd like to. Regardless, we'll give authors a few days to update their repos if they want to keep them in the set, and if that hasn't been done by the end of the weekend then we can push a new package set dropping the now-broken libraries.

chtenb:
@thomashoneyman I'm afraid I don't fully understand what you mean and what's going on. Does this have something to do with transitive dependencies having changed? Since the last release we have dropped no dependencies, only included three more: minibench, partial and transformers.

thomashoneyman:
The issue is that updating string-parsers is causing packages that depend on string-parsers to break. Therefore the new string-parsers can't exist in the package set alongside the packages that depend on it that are no longer compiling.

The solution is to do one of:

- Fix the packages that no longer compile, and update them in this PR so the whole package set builds again
- Remove the packages that no longer compile from the package set, and add them back later
- Don't update string-parsers in the package set

Our default stance is to attempt to fix the packages that don't compile anymore and include the fixed versions in this PR, so the package set has all the same packages. But if it's not possible to update those packages in a short time frame then we remove them from the package set.

chtenb:
Ah I understand @thomashoneyman . I wouldn't mind following up on that, but this is all new to me so I'd need some pointers.

> Would you mind checking if any other packages will be dropped from the set due to this change?

How do I determine which packages are breaking? And how would these packages fix that when string-parsers is not yet in the package set? Should they reference a specific minimum version in spago.dhall or something and reference the string-parsers while bypassing the version from the package set?

thomashoneyman:
> How do I determine which packages are breaking? 

You can clone the package-sets repository at this PR, change directories into `src`, and then run `spago verify-set`. This will reveal packages that are broken.

> And how would these packages fix that when string-parsers is not yet in the package set?

If it's a package using Bower, then it can update the dependency to point at the current `string-parsers` dependency. If it's a package using Spago, it can override its local package set to point at the current `string-parsers` dependency.

chtenb:
The following repositories reference string-parsers. They all need updating, because we removed the Text.Parsing module prefix.

- [ ] https://github.com/drewolson/purescript-biscotti-cookie
- [ ] https://github.com/cdepillabout/purescript-email-validate
- [ ] https://github.com/newlandsvalley/purescript-midi
- [ ] https://github.com/OxfordAbstracts/purescript-graphql-client

chtenb:
@thomashoneyman I noticed that many packages in this package set (include string-parsers) reference an older package-set version themselves, like in https://github.com/purescript-contrib/purescript-string-parsers/blob/main/packages.dhall. Is that a potential issue? Or how does this get resolved?

thomashoneyman:
It's not an issue so long as those libraries are pointing at the updated string-parsers via an override. They may wish to point to an updated package set too, but it's not a problem per se!

chtenb:
@thomashoneyman How would I drop packages from the set? By commenting their entries in the .dhall configuration?

And when would they rejoin the package set? If I understand correctly they have to wait until the next package set release before they can reference the package set that includes string-parsers 7.0.0, is that correct? Or can they already reference the "upcoming" package set?

thomashoneyman:
> How would I drop packages from the set? By commenting their entries in the .dhall configuration?

That's right, you would comment out any broken packages (and any packages that depend on them) until the package set is valid again (ie. `spago verify-set` works).

> And when would they rejoin the package set?

They can rejoin the set at any time. As soon as this PR merges I'll release a new package set that day, and they can update their packages immediately.

>  Can they already reference the "upcoming" package set?

No, they can't reference the 'upcoming' package set, but they can override their local package set to set their `string-parsers` version to v7.0.0.

chtenb:
@thomashoneyman I've commented the packages to drop

src/groups/cdepillabout.dhall
```diff
@@ -1,22 +1,22 @@
...
-  , repo = "https://github.com/cdepillabout/purescript-boxes.git"
-  , version = "v2.1.0"
-  }
-}
+-- { email-validate =
+--   { dependencies = [ "aff", "string-parsers", "transformers" ]
+--   , repo = "https://github.com/cdepillabout/purescript-email-validate.git"
+--   , version = "v6.0.0"
+--   }
+-- , boxes =
```
thomashoneyman, line 6: I wonder why this one is broken? It doesn’t seem to depend on string parsers

src/groups/cdepillabout.dhall
```diff
@@ -1,22 +1,22 @@
...
-    , "newtype"
-    , "profunctor"
-    , "strings"
-    , "tuples"
-    ]
-  , repo = "https://github.com/cdepillabout/purescript-boxes.git"
-  , version = "v2.1.0"
-  }
-}
+-- { email-validate =
```
thomashoneyman, line 1: We can actually just bump this to v7.0.0 - it’s updates

thomashoneyman:
Thanks! Just two quick comments and then we can regenerate the packages.json file and get this in.

src/groups/cdepillabout.dhall
chtenb: Done

src/groups/cdepillabout.dhall
chtenb: I didn't look carefully enough it seems. Fixed again

thomashoneyman:
You’ll need to run “nix run --‘make’” as well and commit the result. Then we’re done!

chtenb:
> You’ll need to run “nix run --‘make’” as well and commit the result. Then we’re done!

Done. And apparently I also needed a `{=}` in the file that became empty.

thomashoneyman (approved):

PR #1086, opened 2022-03-31
Move `httpure` to `citizennet` group

davezuch:
The repo has changed ownership, so we update these groups to reflect that move.


davezuch:
Sorry @thomashoneyman, was traveling. Thanks for handling!

PR #1090, opened 2022-04-21
Add spec, spec-discovery, justifill, fast-vect

sigma-andex:
Add spec, spec-discovery, justifill, fast-vect


sigma-andex:
@JordanMartinez 🏓

JordanMartinez (approved):

PR #1091, opened 2022-04-27
Update purescript ecosystem dependencies

sigma-andex:


JordanMartinez:
Thanks so much @sigma-andex!

PR #1106, opened 2022-05-03
Add some more packages

thomashoneyman:


srid:
@thomashoneyman Is there a reason why some of the packages got commented out in this PR?

I was wondering why @ursi was overriding `html-parser-halogen` [here](https://github.com/purs-nix/purescript-docs-search/blob/2cc174f55e0ef9db28925387ee73e83e271220db/purs.nix#L9).

thomashoneyman:
This PR only uncommented things (the diff is a little noisy). When you see a package commented out in the package sets it’s because at some point it stopped working with the rest of the set. You could try uncommenting it to see if it builds.

This PR specifically was done around the time of the PureScript 0.15 release so there was a lot of churn in working packages.

srid:
Ah, right.

PR #1124, opened 2022-05-09
Add generate-values

JordanMartinez:


src/spago.dhall
```diff
@@ -0,0 +1,228 @@
+{ name = "my-project"
```
thomashoneyman, line 1: What's this about?

JordanMartinez: Oh, whoops! An artifact from when I was testing which packages compile when working on the `prepare-0.15` set

thomashoneyman: Once you take this out and add the source Dhall for `generate-values` I can merge this for ya!

PR #1159, opened 2022-06-02
Update nano-id to v1.1.0

eikooc:


eikooc:
Okay, I am clueless right now on how to fix this issue. It keeps adding dependencies that are not in the project. I can't run the nix docker command locally as it fails with this error:
```
no Makefile, doing nothing
installing
/nix/store/dd81vnld4hfcn7s7i5l4502301ghip6g-stdenv-linux/setup: line 1315:   345 Killed                  "$binPath" --bash-completion-script "$binPath" > "dhall.bash"
error: builder for '/nix/store/1azm61lc6bml3c3ky3la3w06r20vnvrg-dhall-simple.drv' failed with exit code 137
error: build of '/nix/store/1azm61lc6bml3c3ky3la3w06r20vnvrg-dhall-simple.drv', '/nix/store/a5yyqd86p0v5my1657pmanrl30ycmyb1-purescript-v0.15.0.drv', '/nix/store/ipsvp4rqgal6qsghqc0gdcyfdh49phsm-psc-package-simple-0.6.2.drv', '/nix/store/mmbfzbwsgsaficfk5kvxxw5pvljbdkk3-spago-0.20.8.drv', '/nix/store/vcpqjhyqr3ihy5zm4mx8izpjhq552618-dhall-json-simple.drv' failed
```

thomashoneyman:
Is the list of dependencies in the .dhall file correct?

eikooc:
@thomashoneyman I don't know if this is the correct way to check it, but if I `cd src` and run `spago verify-set` it successfully builds all the dependencies in the .dhall file.
I've also run `update-latest-compatible-sets.sh` and `verify-registry.sh` in bash and they run fine without changing anything. I still can't run the nix Docker command, still comes with the same error.

If I run the `make generate` locally in bash I do get the same result as the Github CI action. It adds `console` and `psci-support` to the _packages.json_ file. The repo and tag it is pointing to https://github.com/eikooc/nano-id doesn't have those as dependencies though.

thomashoneyman:
@eikooc the package sets only look at the dependencies listed in the Dhall files, and not the source repository (in this case, the eikooc.dhall file). That’s because a source may not have a spago.dhall file at all (for example, the core libraries) or because many projects include all their development dependencies in their spago file. The files used in this repository are just the dependencies you’d like to go out to users. 

Looking at the eikooc.dhall file under the nano-I’d key, the dependencies list includes console and psci-support. It sounds like you want to delete those, regenerate the packages.json file, and push the result, and then this should work the way you want.

eikooc:
@thomashoneyman oh my god, what a brain fart 🤕 . For some reason I have totally missed the dependencies in that file, I've updated the version there but... anyway. Thanks Thomas!

PR #1192, opened 2022-07-09
Add react-icons v1.0.3

andys8:
Adding my newly created package `react-icons` (PureScript bindings for `react-icons`).

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -3220,6 +3220,15 @@
     "repo": "https://github.com/purescript-contrib/purescript-react-dom.git",
     "version": "v8.0.0"
   },
+  "react-icons": {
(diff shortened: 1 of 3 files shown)


andys8:
Build failed. Potentially because of module size. Looking into it...

https://discordapp.com/channels/864614189094928394/867149806178664478/995309795777187871

thomashoneyman:
Huh — haven't seen this happen before! But we do need this module to compile in order to make it into the package sets. Let's keep this open for the time being.

andys8:
Will only expose the react components instead of helper functions to reduce output size and compile time.
https://github.com/andys8/purescript-react-icons/pull/2

PR #1221, opened 2022-08-27
update js-uri version to v3.1.0

hjmtql:
There was an update about js-uri recently.
https://github.com/purescript-contrib/purescript-js-uri/pull/4


hjmtql:
Is it normal to update packages if a person who is not the maintainers?

thomashoneyman:
Yep! Anyone can update packages!

PR #1222, opened 2022-08-29
add small-ffi

lfarroco:
purescript 0.15-compatible alternative for easy-ffi


thomashoneyman:
You’ll need to register this package in the registry:
https://github.com/purescript/registry 

Once the package is registered with the same URL used here I can restart CI and merge this.

thomashoneyman:
Specifically, add it to this file:
https://github.com/purescript/registry/blob/master/new-packages.json

PR #1226, opened 2022-09-21
Deprecate legacy package sets

thomashoneyman:
See: https://discourse.purescript.org/t/registry-alpha-launched/3146?u=thomashoneyman


thomashoneyman:
I've removed `small-ffi` and `web-page-visibility` from the package sets because they are not in the registry.

f-f (approved):

